PR #12, opened 2019-05-01
Styling & viewport settings layout

marijnvanderhorst:
#### Deployment instructions
Only merge after target branch is development.

#### What have I done
Styled the existing components using a dark theme. Also added the basic structure for the viewport settings component, containing just 3 tabs with mock content.

#### Testing
/


src/_variables.scss
```diff
@@ -3,4 +3,7 @@ $toolbar-height: 4rem;
 /* Override Bootstrap dark theme color to match Oce's */
 $theme-colors: ( // TODO possibly style theme according to oce color palette
   "dark": #1d1d1b
-);
\ No newline at end of file
+);
+
+$body-bg: #313d4c; // Override Bootstrap background color.
+$bg-secondary: #4e5e73; // New variable for secondary background color.
```
Dacaspex: I would not call it "new" because will it be "new" after a while?

Dacaspex: Also, perhaps we should agree on comment styling that we never put comments on the same line? Just a thought open for discussion!

src/app/viewport-settings/viewport-settings.component.html
```diff
@@ -0,0 +1,25 @@
+<ngb-tabset justify="fill">
```
Dacaspex: Perhaps we can put the `viewport-settings` in the `viewport` folder? Like this:
```
viewport/
  viewport-settings/
    viewport-settings.html|ts|scss|spec
  viewport.html|ts|scss|spec
```

src/styles.scss
```diff
@@ -2,3 +2,19 @@
 @import 'variables';
 @import '~bootstrap/scss/bootstrap';
 @import "~@ng-select/ng-select/themes/default.theme.css";
+
+/* Override Bootstrap nav tab styles to work properly with our theme.*/
+.nav-tabs .nav-link.active {
+  color: white;
+  background-color: $bg-secondary;
+  border-bottom: solid $bg-secondary;
```
Dacaspex: This is the same colour as the background, do you want to simulate padding here? Or why is it the same colour?

src/_variables.scss
marijnvanderhorst: New as in "not an overwrite like the other variables in the file", but I agree it's unclear. Will rephrase.

marijnvanderhorst: I agree it's better for consistency and extendability of comment length

src/app/viewport-settings/viewport-settings.component.html
marijnvanderhorst: I don't really agree, as it's really a separate component and not a sub-component of the viewport.

Dacaspex: Fair point, we can leave it like this

src/styles.scss
marijnvanderhorst: Bootstrap's styles need to be overwritten, but I've made some improvements to the overwrite in 36039c5.

src/_variables.scss
```diff
@@ -5,5 +5,8 @@ $theme-colors: ( // TODO possibly style theme according to oce color palette
   "dark": #1d1d1b
 );
 
-$body-bg: #313d4c; // Override Bootstrap background color.
-$bg-secondary: #4e5e73; // New variable for secondary background color.
\ No newline at end of file
+/* Override Bootstrap background color. */
```
Dacaspex: Why `/* comment */` and not `// comment`?

marijnvanderhorst: ¯\_(ツ)_/¯

PR #37, opened 2019-05-06
Importer

Erik9519:
#### Deployment instructions
None

#### What have I done
Import UI / Logic added. Scrum[7].

#### Testing
Click on the Import button in the UI and attempt to load files.


src/app/toolbar/import/import.component.spec.ts
```diff
@@ -0,0 +1,78 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { ImportComponent } from './import.component';
+import {By} from '@angular/platform-browser';
```
Dacaspex: Unused import

src/app/toolbar/import/import.component.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+      spyOn(component.newFilesAvailable, 'emit');
+      spyOn(component.closeImporter, 'emit');
+
+      component.trackFile = component.timingFile = component.modelFile = 'hello';
+      component.onConfirm();
+
+      expect(component.newFilesAvailable.emit).toHaveBeenCalled();
+      expect(component.closeImporter.emit).toHaveBeenCalled();
+  });
+});
```
Dacaspex: Generally looks good. I would add more comments to further explain why you call certain methods (such as the clock for example) and why you test things in certain ways.

Also, the comments / code ratio per file should be > 0.15

src/app/toolbar/import/import.component.ts
```diff
@@ -0,0 +1,59 @@
...
+  templateUrl: './import.component.html',
+  styleUrls: ['./import.component.scss']
+})
+export class ImportComponent implements OnInit {
+    // Files
+    trackFile: string;
+    timingFile: string;
+    modelFile: string;
+    // Events
+    @Output() newFilesAvailable  = new EventEmitter<{trackFile: string, timingFile: string, modelFile: string}>();
```
Dacaspex: I would also properly comment the events, so that people understand what the events are that this component can fire and what they contain. This should also be done in the `/** comment */` syntax like above.

src/app/toolbar/import/import.component.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    ngOnInit() {
+        // Not Required
+    }
+
+    // Retrieve Files from Input[file] fields
+    onFileChange(event, type) {
+        // Instantiate and start File Reader
+        const fr = new FileReader();
```
Dacaspex: Matter of taste, I would avoid using acronyms when the normal variable name would not be that long

src/app/toolbar/import/import.component.html
```diff
@@ -0,0 +1,24 @@
+<div class="container">
+    <h2 class="h2">File Importer</h2>
+    <table>
+        <tr>
+            <label for="trackFile">Track File</label>
+            <input type="file" id="trackFile" accept=".xml"
+                   (change)="onFileChange($event, 'track')">
```
Dacaspex: Code style: Spacing.

I would suggest doing something like this perhaps?
```
<input 
    type="file" 
    id="trackFile" 
    accept=".xml" 
    (change)="onFileChange($event, 'track')"
>
```

src/app/toolbar/import/import.component.html
```diff
@@ -0,0 +1,24 @@
...
+        </tr>
+        <tr>
+            <label for="timingFile">Timing File</label>
+            <input type="file" id="timingFile" accept=".xml"
+                        (change)="onFileChange($event, 'timing')">
```
Dacaspex: Same here

src/app/toolbar/toolbar.component.ts
```diff
@@ -6,11 +6,16 @@ import { Component, OnInit } from '@angular/core';
     styleUrls: ['./toolbar.component.scss']
 })
 export class ToolbarComponent implements OnInit {
+    // Vars
```
Dacaspex: Kinda obvious! :) I would nog add these types of comments, but rather some JSDoc for the `showImporter` variable

Erik9519:
I've done the changes. Should be good now.

PR #38, opened 2019-05-06
Implemented TimeService

TijnBertens:
#### Deployment instructions
/

#### What have I done
Implemented TimeService class., which provides an interface for a global clock. Also implemented unit tests to test individual functions.

#### Testing
Look at unit tests. Add extra ones if necessary.


src/app/backend/time/time.service.ts
```diff
@@ -5,6 +5,135 @@ import { Injectable } from '@angular/core';
...
+
+    /**
+     * Sets the current time to a given timestamp.
+     * @param time The new time.
+     */
+    setCurrentTime(time: number): void {
+        this.currentTime = time;
+
+        // Time can not go below 0
+        this.currentTime = Math.max(0, this.currentTime);
```
marijnvanderhorst: Why not do this max function directly in the assignment above?

src/app/backend/time/time.service.ts
```diff
@@ -5,6 +5,135 @@ import { Injectable } from '@angular/core';
...
+
+            // Update the current time
+            this.currentTime += elapsedSystemTime * this.playbackSpeed;
+
+            // Time can not go below 0
+            this.currentTime = Math.max(0, this.currentTime);
+        }
+
+        // Save the current system time to be the previous system time of the next update
+        this.prevSystemTime = currentSystemTime;
```
marijnvanderhorst: I would put the Max statement here, just to make update robust for negative time.

PR #39, opened 2019-05-06
Implemented Sheet class, added PointOfInterest stub

nolmoonen:
#### Deployment instructions
None.

#### What have I done
Implemented Sheet class and unit tests for Sheet class. Added stub for PointOfInterest since Sheet depends on it.

#### Testing
Run Sheet unit tests.


nolmoonen:
Could someone explain to me why the checks have failed?

Dacaspex:
If a build fails, you can inspect the build logs on Travis. The build on Travis works as follows:
- First, compile the application
- Secondly, run the tests on Firefox and Chrome. You can do this yourself too using the `ng test --watch=false` from your command line
- Lastly, it runs a linter to check your code style. You can do that too using `ng lint` in your command line

If any check fails, the build fails and you get red flags. Let's see what happened here:
image
As we can see, the tests ran fine and everything looks good! (Exit code 0 means everything is fine)

Let's take a look at the linter
image
Oh oh, something went wrong while linting. You can always run the `ng lint` command in your command line too see the output in your console. I can see two things that went wrong, you used `"` instead of `'` and you used the typehint `number`. We have defined to only use `'` for consistency and also agreed that types that can be derived from their value are not needed. 

I think that this should help you fix the issues! :)

src/app/backend/models/sheet/sheet.spec.ts
```diff
@@ -1,7 +1,92 @@
-import { Sheet } from './sheet';
+import {Sheet} from './sheet';
+import {TimeService} from "../../time/time.service";
+import {PointOfInterest} from "../pointofinterest/pointOfInterest";
```
Dacaspex: Did you import the codestyle.xml from the Drive? Imports should have a whitespace between the opening and closing curly braces, like so:
```javascript
import { TimeService } from "..."
```

src/app/backend/models/sheet/sheet.ts
```diff
@@ -1,2 +1,148 @@
...
+
+    /**
+     * Time this sheet starts on the path.
+     */
+    private scheduledMoment: number;
+
+    /**
+     * Last point of interest.
+     */
+    private lastPointOfInterest: PointOfInterest;
```
Dacaspex: Just wondering, should it also have the next point of interest?

src/app/backend/models/sheet/sheet.ts
```diff
@@ -1,2 +1,148 @@
+import {TimeService} from "../../time/time.service";
+import {PointOfInterest} from "../pointofinterest/pointOfInterest";
```
Dacaspex: Code styling just like the other imports

src/app/backend/models/sheet/sheet.ts
```diff
@@ -1,2 +1,148 @@
...
+     */
+    private scheduledMoment: number;
+
+    /**
+     * Last point of interest.
+     */
+    private lastPointOfInterest: PointOfInterest;
+
+    constructor(private timeService: TimeService) {
+        // inject TimeService dependency
```
Dacaspex: No `this.timeService = timeService`? I'm confused?

nolmoonen: I have been informed this is not necessary, nor does it seem that way from testing.

src/app/backend/models/sheet/sheet.ts
nolmoonen: See other comment.

src/app/backend/models/sheet/sheet.ts
nolmoonen: As far as I have been able to find in the requirements, it should not.

src/app/backend/models/sheet/sheet.spec.ts
nolmoonen: No I did not, will do and will fix this.

src/app/backend/models/sheet/sheet.ts
Dacaspex: JavaScript truly is a miracle

src/app/backend/models/sheet/sheet.ts
Dacaspex: 👍

src/app/backend/models/sheet/sheet.spec.ts
nolmoonen: @Dacaspex please add instructions how to import code style to `READE.md`, as well as how to install dependencies. (however, I managed for now)

PR #44, opened 2019-05-07
Sheet service

Erik9519:
#### Deployment instructions
None

#### What have I done
Sheet Service Implementation

#### Testing
Can only be tested with Unit Testing at the moment.


src/app/backend/sheet-service/sheet.service.spec.ts
```diff
@@ -1,12 +1,53 @@
...
+    let sheets: Sheet[];
+
+    beforeEach(() => {
+        TestBed.configureTestingModule({});
+        // Service Instantiation
+        service = new SheetService();
+        // Create Test Sheets
+        sheets = [
+            new Sheet(TestBed.get(TimeService), 0, 0),
+            new Sheet(TestBed.get(TimeService), 4, 1)];
```
Dacaspex: Really minor thing, it would look nicer if the `]` so that you get
```
sheets = [
    new Sheet(...),
    new Sheet(...)
]
```

src/app/backend/sheet-service/sheet.service.ts
```diff
@@ -1,10 +1,62 @@
...
+    getSheetById(id: number): Sheet {
+        return this.sheets.find(s => s.getId() === id);
+    }
+
+    /**
+     * Select a sheet by its ID
+     * @param id represents the id of the sheet you wish to select
+     */
+    setSelectedSheet(id: number) {
+        this.selectedSheet = this.sheets.find(s => s.getId() === id);
```
Extremelyd1: The assignment here uses the same code as the `getSheetById` method, consider calling the method instead.

Issue #25, opened 2019-05-06
[6] Table (UI)


nolmoonen:
Estimate: 3

Dacaspex:
Split up in #124 #125 #126 #127

Issue #24, opened 2019-05-06
[6] Table (logic)


nolmoonen:
Estimate: 6

Dacaspex:
Split up in #124 #125 #126 #127

PR #56, opened 2019-05-09
[do not merge] Feature/table

SquarePlayn:
#### What have I done
Issue [6] Table (UI) #25  
Issue [6] Table (logic) #24 
I created the UI of the table. You can enable/disable columns in the dropdown. Columns can easily be added in the table.component.ts. The data from the sheets is loaded from the SheetService. The function to get the data of a sheet for each specific column can be specified when you add a column using a lambda. When the table is too big for its space, horizontal and/or vertical scrollbars will appear. 
The scrolling depends on the layout restricting the width and height.

#### Testing
Open the dropdown on top of the table. Column names you click should clearly identify to be selected. Selected columns should show in the table. Table columns should remain in the same order after removing and/or adding them. 
When adding the LatPOI row and/or making your screen small, you can make the table wider than its space. When this happens, the table should get a horizontal scrollbar when the width is enforced (e.g. by adding the tag below).
When adding mock sheet data, for example by adding the script below to the ngOnInit() function of table.component.ts, you can make the table have more rows than can fit inside the box vertically. When this is the case, vertical scrollbars should appear. 

Styling to enforce width/height. Will be enforced by layout later. For testing add this to the <div> on line 23 of table.component.html
```
style="height: 45vh; width: 33vw"
```

Script for adding test sheets, meant to be put in the ngOnInit() function of table.component.ts
```typescript
        const sheets: Sheet[] = [];
        for (let i = 0; i < 100; i ++) {
            const sheet = new Sheet(this.timeService, i, 0.3 * i);
            sheet.setVelocity(0.52);
            sheet.setPosition(0.25 * i);
            sheet.setAcceleration(0.5 * i);
            sheet.setLastPointOfInterest(new PointOfInterest());
            sheets.push(sheet);
        }
        this.sheetService.setSheets(sheets);
```
and change the constructor to
```typescript
    constructor(private sheetService: SheetService, private timeService: TimeService) {
    }
```


Dacaspex:
Closing this PR for now as it has become too old.

PR #57, opened 2019-05-09
Feature/sheet multiple timing files

Erik9519:
#### Deployment instructions
Branches in development that use the Sheet class may need to be updated

#### What have I done
Sheet class now supports multiple timing files.

#### Testing
JUnit Testing


src/app/backend/models/sheet/sheet.ts
```diff
@@ -9,34 +9,52 @@ export class Sheet {
...
      */
-    private position: number;
+    private timingID: number[] = [];
 
     /**
-     * Current velocity.
+     * scheduledMoment, position, velocity, acceleration, lastPOI
      */
-    private velocity: number;
+    private data: {scheduledMoment, position, velocity, acceleration, lastPOI}[] = [];
```
SquarePlayn: The data fields can have associated types.

nolmoonen:
* `id` is too ambiguous compared to `timingID`. I propose you change it to `sheetID` and `timingID`, and add more comments to explain the difference.
* Could you add comments to the data structure to explain what all parameters are?

PR #82, opened 2019-05-14
Feature/import layout

Dacaspex:
#### Deployment instructions
/

#### What have I done
- Created the start screen UI and general layout for both desktop and mobile form-factors.
- Removed the old import UI and implementation as it became outdated. It is still available in the history of Github to use as inspiration for the new generation of the import menu. 
- Added routing to provide an easier method for switching between different UIs

#### Testing
Go to `http://localhost:4200/import` on desktop. Observe that you see:
- A picture banner on the left
- A text column on the right
Resize the screen and make sure that the UI does look nice on multiple form-factors. Observe that the UI switches to the mobile version when the screen becomes too small. 

**Note:** The UI for the import selector has not been created. This will be done in a separate story.
**Note:** The UI for the manual import form has not been created. This will be done in a separate story.


src/app/import/import.component.html
```diff
@@ -0,0 +1,124 @@
...
+                    <fa-icon [icon]="arrowLeftIcon" size="2x"></fa-icon>
+                </a>
+            </div>
+        </div>
+
+        <!-- Bottom navigation button, always visible -->
+        <div class="start-container mx-auto px-5">
+            <div class="text-center">
+                <!-- Start button -->
+                <a
```
Extremelyd1: Somehow, this button does not have a pointer cursor

Dacaspex: Fixed

PR #84, opened 2019-05-14
Feature/help layout (Issue #76)

Astrejoe:
#### Deployment instructions
/

#### What have I done
Added responsive help page. Also added routing from and to the workspace.

#### Testing
From workspace, click on the help icon in the toolbar. This should open the help page. Clicking on the 'back to viewport' button should bring you back to the workspace.


src/app/toolbar/help/help-page/help-page.component.scss
```diff
@@ -0,0 +1,23 @@
+@import '../../../../variables';
+
+.help-container {
```
Extremelyd1: This leaves blue bars to the left and right of the text div

src/app/toolbar/help/help-page/help-page.component.scss
```diff
@@ -0,0 +1,23 @@
...
+}
+
+.key-small {
+    width: 50px;
+}
+
+.help-background {
+    width: 100%;
+    height: 100%;
+    background: url('../../../../assets/background/help-bg.jpg');
```
Extremelyd1: This could be shortened to `/assets/background/help-bg.jpg`, since the leading `/` refers to the root directory. This may however lead to a inspection error in WebStorm, which can be ignored.

PR #90, opened 2019-05-14
Data representation of timing file part

SquarePlayn:
#### Deployment instructions
None

#### What have I done
Added classes for data representation of the part of the timing file

#### Testing
Automated testing


src/app/backend/data-representation/sheet-timing.ts
```diff
@@ -0,0 +1,30 @@
+import { TimingEntry } from './timing-entry';
+
+export class SheetTiming {
```
Extremelyd1: Shouldn't this also have a `timingId` to indicate what timing file it is from?

SquarePlayn: Added

PR #97, opened 2019-05-15
Feature/path parser error checking

Erik9519:
#### Deployment instructions
None

#### What have I done
Added pre-processing to path-parser to ensure correct file structure.

#### Testing
JUnit testing


src/app/backend/file-parser/path-parser.ts
```diff
@@ -12,17 +13,138 @@ export class PathParser {
...
+                        'ControlPoint entry ' + c + ' of Shape field in SegmentAppearance of Id ' + segCpId +
+                        ' wrongly formatted.\n' + 'ControlPoint 3rd tag must be \"Name\".'
+                    );
+                }
+            }
+        }
+
+        // ERROR CHECK - Ensure the structure of each Segment entry is Correct
+        for (let i = 0; i < segmentPOIs.length; i++) {
+            // Get the Segment Entry of POIss
```
Chessnl: POIss is used twice in this file and it appears to be a typign error (as I dont see what POIss would stand for), either be more clear on what POIss is or rename to POI's.
```suggestion
            // Get the Segment Entry of POI's
```

src/app/backend/file-parser/path-parser.ts
```diff
@@ -40,45 +162,72 @@ export class PathParser {
...
+        this.findAllPOIs(segmentPOIs, pois, poiInitialized);
 
+        const segments: Segment[] = [];
         // Iterate through each segment to find all control points
         // tslint:disable-next-line:prefer-for-of
         for (let i = 0; i < segmentCPs.length; i++) {
             // Get the Segment Entry of ControlPoints
-            const seg = segmentCPs[i];
+            const segCp = segmentCPs[i].children;
             // Get the Segment Entry of POIss
```
Chessnl: POIss is used twice in this file and it appears to be a typign error (as I dont see what POIss would stand for), either be more clear on what POIss is or rename to POI's.
```suggestion
            // Get the Segment Entry of POI's
```

src/app/backend/file-parser/path-parser.ts
```diff
@@ -12,17 +13,138 @@ export class PathParser {
...
+            if (segCp[1].tagName !== 'Shape') {
+                throw new ParserError(
+                    'SegmentAppearance of Id ' + segCpId + ' must have \"Shape\" as 2nd tag.'
+                );
+            }
+            const shape = segCp[1];
+            // tslint:disable-next-line:prefer-for-of
+            for (let c = 0; c < shape.children.length; c++) {
+                if (c === 1) {
+                    continue;   // Skip Control Point
```
Chessnl: Extend this comment to concisely illustrate why it is ignored (only a suggestion), if this is not possible ignore.

src/app/backend/file-parser/path-parser.ts
```diff
@@ -12,17 +13,138 @@ export class PathParser {
...
+            const segPOIId = segPOI[0].firstChild.nodeValue;
+            if (segPOI[1].tagName !== 'Length') {
+                throw new ParserError(
+                    'Segment of Id ' + segPOIId + ' must have \"Length\" field as 2nd tag.'
+                );
+            }
+            if (segPOI.length > 2) {
+                for (let c = 2; c < segPOI.length; c++) {
+                    if (c === 3) {
+                        continue;   // Skip POI
```
Chessnl: Extend this comment to concisely illustrate why it is ignored (only a suggestion), if this is not possible ignore.

src/app/backend/file-parser/path-parser.ts
```diff
@@ -40,45 +162,72 @@ export class PathParser {
...
+            const poisInSegment: POI[] = [];
+
             // Store all control points
-            const cps = seg.children[1];
+            const shape = segCp[1];
             // tslint:disable-next-line:prefer-for-of
-            for (let c = 0; c < cps.children.length; c++) {
+            for (let c = 0; c < shape.children.length; c++) {
                 if (c === 1) {
                     continue;   // Skip Control Point
```
Chessnl: Extend this comment to concisely illustrate why it is ignored (only a suggestion), if this is not possible ignore.

Chessnl: Extend this comment to concisely illustrate why it is ignored (only a suggestion), if this is not possible ignore.

src/app/backend/file-parser/path-parser.ts
```diff
@@ -91,12 +240,33 @@ export class PathParser {
                 if (p === 3) {
                     continue;   // Skip POI
```
Chessnl: Extend this comment to concisely illustrate why it is ignored (only a suggestion), if this is not possible ignore.

src/app/backend/file-parser/path-parser.ts
```diff
@@ -12,17 +13,138 @@ export class PathParser {
...
+                    }
+                }
+            }
+        }
+
+        // ERROR CHECK - Ensure that SegmentAppearance entries appear in the same order as Segment entries
+        for (let i = 0; i < segmentCPs.length; i++) {
+            if (segmentCPs[i].children[0].firstChild.nodeValue !== segmentPOIs[i].children[0].firstChild.nodeValue) {
+                throw new ParserError(
+                    'Missmatch of Id\'s / ordering between SegmentAppearance and Segment.\n' +
```
Chessnl:
```suggestion
                    'Mismatch of Id\'s / ordering between SegmentAppearance and Segment.\n' +
```

src/app/backend/file-parser/path-parser.spec.ts
```diff
@@ -130,4 +164,224 @@ describe('PathParser', () => {
...
+            'Not all POI\'s found in the SegmentAppearance of Id 2 have been found in the respective Segment.')
+        );
+    });
+
+    it ('should throw error if POI is defined in Segment but not in SegmentAppearance', () => {
+        template.splice(12, 0, '<Segment>\n');
+        template.splice(13, 0, '<Id>2</Id>\n');
+        template.splice(14, 0, '<Length>341</Length>\n');
+        template.splice(15, 0, '<PointOfInterest>\n');
+        template.splice(16, 0, '<Type>lol</Type>\n');
```
Chessnl: I would consider 'lol' a bit unprofessional :-) maybe use 'fake' or something.

src/app/backend/file-parser/path-parser.ts
Dacaspex: Shouldn't it then be `POIs`?

src/app/backend/file-parser/path-parser.ts
Dacaspex: See my other comment, shouldn't this become `POIs`?

src/app/backend/file-parser/path-parser.spec.ts
```diff
@@ -130,4 +164,224 @@ describe('PathParser', () => {
         expect(poiPosTest1).toBe(interfacePoint6);
         expect(poiPosTest2).toBe(interfacePoint6);
     });
+
+    /**
```
Dacaspex: Small thing, `/** */` are used for function and class documentation blocks. Minor change to `/* */`

src/app/backend/file-parser/path-parser.ts
Dacaspex: Important comment for maintenance

src/app/backend/file-parser/path-parser.ts
```diff
@@ -12,17 +13,138 @@ export class PathParser {
...
+                );
+            }
+            const segCpId = segCp[0].firstChild.nodeValue;
+            if (segCp[1].tagName !== 'Shape') {
+                throw new ParserError(
+                    'SegmentAppearance of Id ' + segCpId + ' must have \"Shape\" as 2nd tag.'
+                );
+            }
+            const shape = segCp[1];
+            // tslint:disable-next-line:prefer-for-of
```
Dacaspex: Can't this be disabled in the tslint config? I am not a fan of meta information in comments

src/app/backend/file-parser/path-parser.ts
```diff
@@ -12,17 +13,138 @@ export class PathParser {
...
     constructor(private path: Path) {}
 
-    parse(txt: string) {
-        const segments: Segment[] = [];
+    /**
+     * This function ensures that the structure of the file is correct. It does not look at the values themselves other
+     * than to help the User locate where in the {@code txt} a structure is found.
+     * @param txt contains the text to parse
+     */
+    private preParse(txt: string) {
```
Dacaspex: Can and should be made static in my opinion

src/app/backend/file-parser/path-parser.ts
```diff
@@ -12,17 +13,138 @@ export class PathParser {
...
+        }
+
+        // ERROR CHECK - Ensure the structure of each SegmentAppearance entry is Correct
+        for (let i = 0; i < segmentCPs.length; i++) {
+            // Get the Segment Entry of ControlPoints
+            const segCp = segmentCPs[i].children;
+
+            if (segCp[0].tagName !== 'Id') {
+                throw new ParserError(
+                    'SegmentAppearance entry number ' + i + ' must have \"Id\" as 1st tag.'
```
Dacaspex: Why are `"` escaped? This is not necessary since the outer quotation marks are `''` and not `""`. This is done on multiple places

src/app/backend/file-parser/path-parser.ts
```diff
@@ -12,17 +13,138 @@ export class PathParser {
...
+                );
+            }
+            const segCpId = segCp[0].firstChild.nodeValue;
+            if (segCp[1].tagName !== 'Shape') {
+                throw new ParserError(
+                    'SegmentAppearance of Id ' + segCpId + ' must have \"Shape\" as 2nd tag.'
```
Dacaspex: See escaping comment

src/app/backend/file-parser/path-parser.spec.ts
```diff
@@ -130,4 +164,224 @@ describe('PathParser', () => {
...
+     * PRE-PARSING CHECKs
+     */
+
+    it('should throw missmatch error of the number of Segment vs SegmentAppearance entries', () => {
+        template.splice(12, 0, '    <Segment>\n');
+        template.splice(13, 0, '      <Id>2</Id>\n');
+        template.splice(14, 0, '      <Length>341</Length>\n');
+        template.splice(15, 0, '    </Segment>\n');
```
Extremelyd1: It's hard to see from these lines of code what is actually happening to the data that is supposed to be parsed. When viewing this test case I can't immediately tell why it is supposed to throw an error. Perhaps create multiple datasets, which you use in the different test cases.

This comment applies to a lot of test cases in this file.

src/app/backend/file-parser/path-parser.spec.ts
```diff
@@ -130,4 +164,224 @@ describe('PathParser', () => {
...
+        const path = new Path([], []);
+        const parser = new PathParser(path);
+        expect(() => { parser.parse(template.join('')); }).toThrow(new ParserError(
+            'Missmatch of Id\'s / ordering between SegmentAppearance and Segment.\n' +
+            'SegmentAppearance of Id 16 was matched with Segment of Id 1.')
+        );
+    });
+
+    it ('should throw error if there are multiple coordinate definitions for a POI', () => {
+        template.splice(12, 0, '    <Segment>\n');
```
Extremelyd1: Same as other comment, you edit a lot of the xml data here, but it is unclear what happens and why this should throw an error.

src/app/backend/file-parser/path-parser.ts
```diff
@@ -12,17 +13,138 @@ export class PathParser {
...
+        for (let i = 0; i < segmentCPs.length; i++) {
+            // Get the Segment Entry of ControlPoints
+            const segCp = segmentCPs[i].children;
+
+            if (segCp[0].tagName !== 'Id') {
+                throw new ParserError(
+                    'SegmentAppearance entry number ' + i + ' must have \"Id\" as 1st tag.'
+                );
+            }
+            const segCpId = segCp[0].firstChild.nodeValue;
```
Dacaspex: Using short names becomes very difficult to read, especially when you are not very familiar with the code base. Here the naming does not properly reflect the property as well. `segCpId` implies `segment control point id`, i.e. the id of a control point, but this is actually the id of a segment (which contains multiple control points)

src/app/backend/file-parser/path-parser.ts
```diff
@@ -12,17 +13,138 @@ export class PathParser {
...
         const segmentCPs  = new DOMParser().parseFromString(txt, 'text/xml').getElementsByTagName('SegmentAppearance');
         // DOMParse for Points Of Interest
         const segmentPOIs = new DOMParser().parseFromString(txt, 'text/xml').getElementsByTagName('Segment');
 
-        // Used to keep of whether or not a POI has been fully initialized
-        const poiInitialized: boolean[] = [];
-        // These are all the POI's that exist in the path
-        const pois: POI[] = [];
+        // ERROR CHECK - Ensure the number of segments matches the number of SegmentAppearances
+        if (segmentCPs.length !== segmentPOIs.length) {
```
Extremelyd1: `segmentCP` is not very descriptive, maybe rename it to `segmentControlPoint`.

src/app/backend/file-parser/path-parser.ts
```diff
@@ -12,17 +13,138 @@ export class PathParser {
...
+                'The number of SegmentAppearance entries must be equal to the number of Segment entries.\n' +
+                'There are ' + segmentCPs.length + ' SegmentAppearance Entries whilst there are ' +
+                segmentPOIs.length + ' Segment entries.'
+            );
+        }
+
+        // ERROR CHECK - Ensure the structure of each SegmentAppearance entry is Correct
+        for (let i = 0; i < segmentCPs.length; i++) {
+            // Get the Segment Entry of ControlPoints
+            const segCp = segmentCPs[i].children;
```
Extremelyd1: Same as other comment, `segCP` is not very descriptive.

src/app/backend/file-parser/path-parser.ts
```diff
@@ -12,17 +13,138 @@ export class PathParser {
...
+                if (c === 1) {
+                    continue;   // Skip Control Point
+                }
+                if (shape.children[c].tagName !== 'ControlPoint') {
+                    throw new ParserError(
+                        'Shape field in SegmentAppearance of Id ' + segCpId + ' wrongly structured.\n' +
+                        'Shape field must contain a list of ControlPoint entries.'
+                    );
+                }
+                if (shape.children[c].children[0].tagName !== 'xCoordinate') {
```
Dacaspex: This becomes very difficult to read as it is two layers of children with little context about what is what. It took me quite some effort to understand what is what

src/app/backend/file-parser/path-parser.ts
```diff
@@ -12,17 +13,138 @@ export class PathParser {
...
+                );
+            }
+            const segPOIId = segPOI[0].firstChild.nodeValue;
+            if (segPOI[1].tagName !== 'Length') {
+                throw new ParserError(
+                    'Segment of Id ' + segPOIId + ' must have \"Length\" field as 2nd tag.'
+                );
+            }
+            if (segPOI.length > 2) {
+                for (let c = 2; c < segPOI.length; c++) {
```
Extremelyd1: Not necessary to change, but I find `c` to be a weird loop variable.

src/app/backend/file-parser/path-parser.ts
Dacaspex: Also, you specifically check the order of the xml tags. This should not be necessary as all tags are named. I don't think it is a good idea to assume the same ordering of the xml files

src/app/backend/file-parser/path-parser.ts
```diff
@@ -91,12 +240,33 @@ export class PathParser {
...
+                        'respective SegmentAppearance.'
+                    );
                 }
             }
+            // ERROR CHECK - Ensure that all POI's found in this Segment are specified in the respective
+            //               SegmentAppearance
+            if (poisInSegment.length > 0) {
+                throw new ParserError(
+                    'Not all POI\'s found in the SegmentAppearance of Id ' + segId + ' have been found in the ' +
+                    'respective Segment.'
```
Extremelyd1: Not something you should change, but the above 3 lines had me confused for a minute. Github at its finest.

src/app/backend/file-parser/path-parser.ts
Erik9519: Not sure how do disable it in the config.

Dacaspex: Opened a new issue for that #116

Issue #122, opened 2019-05-17
Toggleable panels have issues with 100% height


Erik9519:
UI modules placed in the toggle-able panels box cannot rely on using 100% height scaling as they otherwise ignore the box boundaries.

cwognum:
This was no actual bug, rather a mistake in the Settings module

PR #128, opened 2019-05-17
Feature/settings layout

ArtixOfVenion:
#### Deployment instructions
\

#### What have I done
Added responsive settings configuration panel, with dynamic item listing.
Panel can be viewed from either the toggleable workspace panels or through the mobile link at the top right. (Mobile view still has an issue with sizing which was fixed for desktop).

#### Testing
Go to the workspace and open the config panel and notice responsive list and tabs. In the 'settings.component.ts' file there are dummy variables which will show the styling and responsiveness of list items.
'lol' - Jeroen


src/app/settings/settings.component.scss
```diff
@@ -0,0 +1,131 @@
...
+
+.toolbar-btn {
+    vertical-align: middle;
+}
+
+.text-color {
+    color: black;
+}
+
+@media (min-width: 576px) { // ToDo Use BootStrap sm
```
Extremelyd1: TODO: fix todo. :)

src/app/settings/settings.component.html
```diff
@@ -0,0 +1,64 @@
...
+                type="button"
+                class="btn btn-secondary toolbar-btn">
+                <fa-icon [icon]="sheetIcon"></fa-icon>
+                <span class="d-none d-lg-inline-block ml-1">Sheet</span>
+            </button>
+        </div>
+    </div>
+
+    <div class="background">
+        <div class="" style="overflow-y: scroll; height: calc(20em - 45px);"> <!-- TODO: Change to the class, replace hardcoded 20em with 100% -->
```
cwognum: - I would not add inline CSS, but keep HTML and CSS completely separate
- For `overflow-y: scroll` we defined a CSS class in styles.scss called .overflow-y-scroll 
- I would change the hard coded value in the height calculations by SASS variables for now. Ideally we would be able to change this to 100%, but I am not sure how to do so (see also #122)

src/app/settings/settings.component.html
```diff
@@ -0,0 +1,64 @@
...
+                (click)="this.selectedTabIndex = 2;"
+                type="button"
+                class="btn btn-secondary toolbar-btn">
+                <fa-icon [icon]="sheetIcon"></fa-icon>
+                <span class="d-none d-lg-inline-block ml-1">Sheet</span>
+            </button>
+        </div>
+    </div>
+
+    <div class="background">
```
cwognum: For the sake of clarity and to avoid naming collisions, I would change this class name to something a bit more specific. For example; `settings-background`

src/app/settings/settings.component.html
```diff
@@ -0,0 +1,64 @@
...
+        <div class="" style="overflow-y: scroll; height: calc(20em - 45px);"> <!-- TODO: Change to the class, replace hardcoded 20em with 100% -->
+            <div *ngFor="let item of items[this.selectedTabIndex]">
+                <div class="d-flex flex-row w-100 settings-row my-1 py-1" style="align-items: center">
+                    <label class="mr-auto pl-2 text-color">
+                        {{item.name}}
+                    </label>
+                    <div class="pr-4">
+                        <label class="item" *ngIf="(item.type == 'boolean')">
+                            <div class="container-checkmark">
+                                <input type="checkbox" [(ngModel)]="item.value">
```
cwognum: To make the different setting types more easily extendable, we could only allow the HTML types for the item.type variable and generate the settings like this, instead of different NgIf statements: 

``` HTML
<label class="item">
    <input type="{{item.type}}" [(ngModel)]="item.value">
</label>
```

src/app/settings/settings.component.html
```diff
@@ -0,0 +1,64 @@
...
+                class="btn btn-secondary toolbar-btn">
+                <fa-icon [icon]="sheetIcon"></fa-icon>
+                <span class="d-none d-lg-inline-block ml-1">Sheet</span>
+            </button>
+        </div>
+    </div>
+
+    <div class="background">
+        <div class="" style="overflow-y: scroll; height: calc(20em - 45px);"> <!-- TODO: Change to the class, replace hardcoded 20em with 100% -->
+            <div *ngFor="let item of items[this.selectedTabIndex]">
```
cwognum: Suggestion: Consistently rename 'item' to 'setting' for clarity?

src/app/settings/settings.component.html
```diff
@@ -0,0 +1,64 @@
...
+                <div class="d-flex flex-row w-100 settings-row my-1 py-1" style="align-items: center">
```
cwognum: Replace inline CSS by the bootstrap class: `.text-center`

src/app/settings/settings.component.scss
```diff
@@ -0,0 +1,131 @@
...
+.settings-row {
+    background-color: white;
+}
+
+.background {
+    background-color: #EBEBEB;
+}
+
+.toolbar-btn {
+    vertical-align: middle;
```
cwognum: There already is a Bootstrap class for this: `.align-middle`

src/app/settings/settings.component.ts
```diff
@@ -0,0 +1,55 @@
...
+        // new SettingsEntry('Drum', true, 'boolean'),
+        // new SettingsEntry('Printer head', true, 'boolean'),
+    ], [
+        // Sheet Tab Settings
+        // new SettingsEntry('Show sheet ID', true, 'boolean'),
+        // new SettingsEntry('Sheet Color', '#FFFFFF', 'color'),
+    ]];
+    private selectedTabIndex = 0;
+
+    private filmIcon = faFilm;
```
cwognum: Suggestion: Name the icons for what they are used (`sceneIcon`), not for how they look (`filmIcon`). If we then are to update the icon later on by a different icon, the alias is still correct.

src/app/settings/settings.component.ts
```diff
@@ -0,0 +1,55 @@
...
+    private cubeIcon = faCube;
```
cwognum: Same as above

src/app/settings/settings.component.scss
```diff
@@ -0,0 +1,131 @@
...
+    border-bottom-color: $oce-red;
+    background-color: white;
+    color: $oce-red;
+}
+.deselected {
+    background-color: white;
+    color: grey;
+}
+
+.settings-row {
```
cwognum: There is a Bootstrap class for this already: `.bg-white`

cwognum:
> 'lol' - Jeroen

I would consider 'lol' a bit unprofessional @ArtixOfVenion

src/app/settings/settings.component.html
ArtixOfVenion: I resolved all comments except for the one relating to #122 , since we were not able to fix that hence the issue.

src/app/settings/settings.component.html
ArtixOfVenion: This did not seem to give the same result so I left the css but put it in a css class.

src/app/settings/settings.component.scss
ArtixOfVenion: Since the settings row has a mediaquery css, setting the bg-white in the class would override our mediaquery so I left it as it is for now.

src/app/settings/settings.component.html
Erik9519: I've attempted to do what you suggested however for some reason the second that the type of the input is set by a variable every time I change tab it forgets what value it previously had. It does not happen if it is set straight up with text.

```
<label class="setting">
   <div class="container-{{item.type}}">
      <input type="{{item.type}}" [(ngModel)]="item.value">
     <span class="span-{{item.type}}"></span>
  </div>
</label>
```

src/app/settings/settings.component.scss
Erik9519: I like ToDo better but ok :)

src/app/settings/settings.component.html
cwognum: You can still change it to variables. I would not use hard coded values. I of course understand that it can't be changed to the 100% unit at the moment :)

src/app/settings/settings.component.html
cwognum: Hm. I'll have a look at it tonight to see if I can figure it out. Otherwise we can leave it like this! :)

src/app/settings/settings.component.html
cwognum: Oh yeah. I read that wrong! I thought it was a text align.

src/app/settings/settings.component.scss
cwognum: Ah, okay! Makes sense!

src/app/settings/settings.component.scss
cwognum: I think Yannick meant that the TODO should be implemented before merging the PR

src/_variables.scss
```diff
@@ -18,3 +18,8 @@ $bg-secondary: #4e5e73;
 
 /* Highlight color. */
 $highlight: #e5e5e5;
+
+/* Config container toolbar height*/
+$config-container-toolbar-height: 45px;
```
cwognum: Config? Settings? See also #134

cwognum:
I noticed that you changed the hardcoded units in the `calc()` statement to variables. However, you did not seem to change the original values to variables as well. The 20em you use is the fixed height of the toggle panels on Desktop. Therefore, you should also change the height of the toggle panels in `desktop-toggle-panels.component.scss`: 

``` css
.toggle-panel {
    height: $config-container-height;
}
```
The same holds for the 45px, although I do not know where that value came from as it only appears once in the entire project (where you create the variable)

image

Oh and try to be consistent with config(uration) or settings. This is a larger issue that is not to be fixed within the scope of this PR (also see #134), but just be aware because some mixed terminology is used. 
image

src/app/settings/settings.component.html
cwognum: I did not get to it unfortunately. Let's discuss this on monday! :)

cwognum:
Looks good! I would like to add a redirect / conditional routing when on desktop to make the settings page unavailable at that point. Perhaps something like [this](https://stackoverflow.com/questions/47133610/angular-4-manual-redirect-to-route)?

Extremelyd1:
The change I requested with the title in the settings menu will be addressed in #145.

Issue #130, opened 2019-05-17
Make timing parser more robust for input files where time goes backwards


Dacaspex:

marijnvanderhorst:
Solved in PR #135.

PR #135, opened 2019-05-17
#130 Make timing parser robust against negative time difference

SquarePlayn:
#### What have I done
Made the timing parser robust against timing data in which two consecutive timing entries of one sheet have a negative time difference. 
In other words, when the time goes backward at some point in the timing data, the parser will throw a ParseError. 

#### Testing
Automated testing


src/app/backend/file-parser/timing-parser.service.ts
```diff
@@ -100,6 +101,15 @@ export class TimingParserService {
                 const velocity = Number(TimingParserService.getTagValue(timingEntryXml, 'Velocity'));
                 const acceleration = Number(TimingParserService.getTagValue(timingEntryXml, 'Acceleration'));
 
+                // Check for time going backwards
+                if (lastTime !== undefined) {
```
marijnvanderhorst: Maybe initialize `lastTime` with a negative value, as we can safely assume the timing file will never contain negative timestamps.

marijnvanderhorst: `Number.NEGATIVE_INFINITY` might be a better solution now I think of it.

src/app/backend/file-parser/timing-parser.service.ts
```diff
@@ -100,6 +101,15 @@ export class TimingParserService {
...
+                    const timeDif = time - lastTime;
+                    if (timeDif < 0) {
```
marijnvanderhorst: Why not just `time < lastTime`?

src/app/backend/file-parser/timing-parser.service.ts
Dacaspex: But then the file could contain negative timestamps? Perhaps we should write a different parser test for that

src/app/backend/file-parser/timing-parser.service.ts
SquarePlayn: Updated

src/app/backend/file-parser/timing-parser.service.ts
SquarePlayn: It is easy to add a check to see whether there are no negative time entries but currently I have not heard the people from the sheet state engine calculation make/need that assumption so I decided not to put it in. @marijnvanderhorst if you think it is needed for the engine for time to be positive (or perhaps additionally for it to start at 0.000 at the first entry), let me know and I'll add that.

Dacaspex: I approve the changes as long as it is fine to leave out the negative time check

marijnvanderhorst: Currently I don't think we need the assumption of positive timing entries. It is also not the scope of this issue. Changes look good, can be merged imo.

PR #181, opened 2019-05-21
Added route positions for all pois on route

nolmoonen:
#### Deployment instructions
None.

#### What have I done
Each POI on the path will have an array containing its position(s) on the route.

#### Testing
Run the single added test case.


src/app/backend/data-representation/poi.ts
```diff
@@ -6,14 +6,34 @@ export class POI extends ControlPoint {
      */
     private name: string;
 
+    /**
+     * Array of positions on the route.
```
marijnvanderhorst: "Array of positions where this POI occurs on the route."

nolmoonen: Changed.

PR #197, opened 2019-05-22
Feature/get num timings

Chessnl:
#### Deployment instructions
-

#### What have I done
The table UI depends on whether 1 or more timing models are loaded in. As for this component to know how many are loaded it, it will ask the sheetService. The sheetEngine now forwards the numTimings from the sheetContainer to the sheetService.

#### Testing
-


src/app/backend/sheet-state-engine/sheet-state-engine.service.ts
```diff
@@ -36,8 +36,12 @@ export class SheetStateEngineService {
             const sheet: Sheet = sheets[si];
             const sheetTimingStates: SheetTimingState[] = [];
 
+            // Forwards the number of loaded in timing models to the sheetService
+            const numTimings: number = this.sheetContainer.getNumTimings();
```
SquarePlayn: Should move outside of the for loop

Chessnl: Incorporated

PR #199, opened 2019-05-22
Feature/graph

ArtixOfVenion:
#### Deployment instructions
/

#### What have I done
Added a responsive graph to the desktop and mobile toggle panels.

#### Testing
Open Exhibit and toggle the panels for the graph and check whether the other panels when toggled resize accordingly. Go to mobile version and check whether you can swipe to the graph panel and whether it is responsive.


src/app/graph/graph.component.html
```diff
@@ -1,9 +1,13 @@
...
-<!--    <canvas baseChart-->
-<!--            [datasets]="chartData"-->
-<!--            [options]="chartOptions"-->
-<!--            [legend]="chartLegend"-->
-<!--            [chartType]="chartType"-->
-<!--    >-->
-<!--    </canvas>-->
+<div class="d-flex w-100 h-100 flex-column bg-dark">
+    <app-graph-config class="config"></app-graph-config>
+    <div class="w-100 chart" style="position: relative">
```
Vogeltak: If you already have a class in css, you should rather also write this in there, instead of the current in-line style option.

PR #201, opened 2019-05-22
Graph configuration

Vogeltak:
#### Deployment instructions
It should be included in the Graph component html in order for it to actually be used.

#### What have I done
The graph configuration adds an ng-select component that enables a user to select a single graph metric that is to be plotted in the graph itself.

#### Testing
Run unit tests. Note that the test for `onChange()` has not been written yet, because it has no meaningful implementation at this point.


src/app/graph/graph-config/graph-config.component.ts
```diff
@@ -0,0 +1,29 @@
...
+    ngOnInit() {
+    }
+
+    /**
+     * Function to be called whenever the selected graph metric is changed.
+     * It should connect to the Graph service in order to connect the correct data to the graph.
+     * @param event String value of the selected option.
+     */
+    onChange(event: {}) {
+        console.log(event);
```
ArtixOfVenion: Replace with ToDo

src/app/graph/graph-config/graph-config.component.html
```diff
@@ -0,0 +1,5 @@
+<ng-select [items]="metricLabelNames"
```
ArtixOfVenion: I don't mind but I believe most people want to have the closing tag on a newline

PR #208, opened 2019-05-22
Updated outdated package lock.

marijnvanderhorst:
#### Deployment instructions
/

#### What have I done
Updated the `package-lock.json` file as it was not pushed when the contents of the `package.json` were updated in previous PRs.

#### Testing
After checkout execute `npm install` and observe that the `package-lock.json` does not change.


marijnvanderhorst:
> For me running `npm install` does give a different `package-lock.json`. It adds many `optional: true` fields. npm version 6.9.0

Good catch! I was using an older version of npm apparently.

Issue #209, opened 2019-05-23
Path Parser Coordinate System Scaling


Erik9519:
The Path parser currently does not reverse the Y coordinate, doesn't divide the coordinate by 1000 and it doesn't offset the entire path by the lowest control point found.

PR #210, opened 2019-05-23
#209 Implemented proper coordinate offseting and scaling

Erik9519:
#### Deployment instructions
None

#### What have I done
Implemented the control points Y coordinate flipping, division of coordinates by 1000 and offsetting of the path by the lowest control point found as done in Argus3D.

#### Testing
JUnit testing.


src/app/backend/file-parser/path-parser.service.ts
```diff
@@ -83,6 +85,12 @@ export class PathParserService {
                             segmentAppearanceId + ' wrongly formatted.\n' + 'ControlPoint must contain one Name tag.'
                         );
                     }
+                    // Keep track of lowest point in the path
+                    const cpY = Number(controlPoint[c].getElementsByTagName('yCoordinate')[0].childNodes[0].nodeValue)
```
Dacaspex: What does `cpY` stand for? I would rename this variable to something more descriptive

Erik9519: cpY stands for Control Point Y Coordinate.

I would've created a separate function for it but that would mean looping through the entire file another time just to gather that data. (Due to the immutable nature of our Data structures)
Felt like too much of a performance waste.

Also not doing the scaling at the source, meaning editing the Path object itself, I feel like would add an extra layer of complexity which is not needed.
It can be done at rendering time, then again, this is a problem that shouldn't exist to begin with as especially moving upwards the path by the amount of the lowest point seems a very arbitrary thing to do.

Personally I would keep this processing at the source.

src/app/backend/models/printer-model.ts
```diff
@@ -0,0 +1,61 @@
+import { Injectable } from '@angular/core';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class PrinterModel {
+    private lowestControlPointX: number;
```
Dacaspex: Add documentation

src/app/backend/file-parser/path-parser.service.ts
```diff
@@ -227,11 +224,14 @@ export class PathParserService {
...
                 // Get Control Point Coordinates
                 const cp = new ControlPoint(
-                    Number(controlPoint[c].getElementsByTagName('xCoordinate')[0].childNodes[0].nodeValue)
-                    /  1000,
+                    Number(controlPoint[c].getElementsByTagName('xCoordinate')[0].childNodes[0].nodeValue),
                     Number(controlPoint[c].getElementsByTagName('yCoordinate')[0].childNodes[0].nodeValue)
-                    / -1000 - lowestPointOnPath
                 );
+                // Check Boundaries
+                if (cp.getX() < lowestControlPointX ) { lowestControlPointX  = cp.getX(); }
```
Dacaspex: I would put these on separate lines

src/app/backend/models/printer-model.ts
```diff
@@ -0,0 +1,61 @@
...
+     * @param lowestControlPointX  lowest X coordinate
+     * @param highestControlPointX highest X coordinate
+     * @param lowestControlPointY  lowest Y coordinate
+     * @param highestControlPointY highest Y coordinate
+     */
+    setControlPointsBoundaries(
+        lowestControlPointX: number,
+        highestControlPointX: number,
+        lowestControlPointY: number,
+        highestControlPointY: number): void {
```
Dacaspex: I would put the `): void {` on a separate line

src/app/backend/file-parser/path-parser.service.ts
Erik9519: Code overhaul

src/app/backend/models/printer-model.ts
Erik9519: Added

src/app/backend/file-parser/path-parser.service.ts
Erik9519: Changed

src/app/backend/models/printer-model.ts
Erik9519: Changed

tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-    "compileOnSave": false,
+    "compileOnSave": true,
```
Dacaspex: Should be set back to false

Issue #212, opened 2019-05-23
Hide content when collapsing a panel


cwognum:
![image](https://user-images.githubusercontent.com/28000055/58237055-8e4a6880-7d44-11e9-8806-0c2d4fa8aede.png)

When fully collapsing a panel on mobile, the content is still visible.

Issue #204, opened 2019-05-22
Mobile slider panels on expansion and collapse redraw canvas


ArtixOfVenion:
Currently when the graph canvas is drawn on screen and the slider panels are extended the graph will stretch. The arrows should probably have some redraw on extending the panels.

Issue #203, opened 2019-05-22
Mobile graph panel slider at expansion level 2 gets overlapped


ArtixOfVenion:
The downwards arrow gets overlapped by the graph config dropdown menu.

PR #213, opened 2019-05-23
#212, #204 & #203 Fixed all issues

ArtixOfVenion:
#### Deployment instructions
/

#### What have I done
#212 #204 #203 
#### Testing
Test visually, by opening the graph on mobile and collapsing and extending it. When it is extended all the way it should not stretch and it should not overlap with the collapse button. When it is collapsed all the way it should not show the content of the graph panel.


cwognum:
@nolmoonen I thought we still had to update the test cases for these?

Issue #224, opened 2019-05-23
Feedback to the user that the parsing failed


Dacaspex:
As of now, when the parsing fails in the import screen, the error is logged to the console.

Preferably, the user should be informed (by means of a pop up, notification, etc. ) that something has gone wrong and what has gone wrong.

Issue #223, opened 2019-05-23
Import second timing file


Dacaspex:
Currently, the import component does not have functionality to import a second timing file. 

Also, the parse service does not yet have this functionality.

PR #225, opened 2019-05-23
Feature/parse service

### LINKED ISSUE #149: Call file parsers from import menu when files are imported
At the moment, the files are only put into the `FileStorage`. However, the import component should also call the parsers to actually parse the files into the logical model. 

I think the import component is suitable for this task, since it can then also directly show the errors on the import screen.

Dacaspex:
#### Deployment instructions
/

#### What have I done
fixes #149 
- Created the parse services that calls all file parses in order. 
- Created function to import a track, path and timing data from the front-end
    - Did not yet implement the more elaborate function to import more timing files, see #223 
    - Did not yet implement the functionality to import the 3D model, see #215 
- Hooked up import button with back-end.
- Changed implementation of parsers and dependent classes to be proper singletons without any parameters in the constructor call.

#### Testing
**Note:** The unit tests do not work because of a bug in Angular zone.js, see comment in test cases.
**Note:** This causes the code coverage to not be 100%, unfortunately

1. Go to the import screen `localhost:4200/import` and navigate to the import manually page (pressing the plus icon).
2. Select some files from Océ 
3. Observe the output in the console (issue #224 will implement an actual UI component for this error)


Dacaspex:
Fixes #149

PR #232, opened 2019-05-24
Feature/Setup Viewport Component

### LINKED ISSUE #214: [2] Setup Viewport Component

### LINKED ISSUE #233: Firefox does not support webgl in headless mode
Relates to #214

Erik9519:
#### Deployment instructions
None

#### What have I done
Setup of the view-port (Disregard the name of the branch as it is wrong)

#### Testing
Junit testing for the dispatcher. Try Exhibit on your computer and see that the canvas is properly sized.

fixes #233 
fixes #214


Dacaspex:
We will test this with the new Travis configuration on Monday, see #234

src/app/viewport/viewport.component.ts
```diff
@@ -1,16 +1,117 @@
...
+    /**
+     * ThreeJS Camera
+     */
+    private camera: THREE.Camera;
+    /**
+     * Stats.js component
+     */
+    private stats: STATS;
+
+    constructor(private dispatcher: DispatcherService) {
```
marijnvanderhorst: This introduces a two way dependency between the Dispatcher and the Viewport, which is not desired. I see why you cannot have dispatch depend on viewport, so maybe turn the dependency around.

src/app/viewport/viewport.component.ts
```diff
@@ -1,16 +1,117 @@
...
+    }
+
+    /**
+     * Initializes the Stats component
+     */
+    private initializeStats(): void {
+        this.stats = new STATS();
+        // Change the default stats style to absolute position (instead of relative) and z-index to 1
+        // (instead of 10000)
+        this.stats.dom.style.cssText = 'position:absolute;top:0;left:0;cursor:pointer;opacity:0.9;z-index:0';
```
marijnvanderhorst: Comment says 1, but code says 0. Also, please add a comment why this is needed.

src/app/viewport/viewport.component.html
```diff
@@ -1,3 +1,3 @@
-<div class="viewport-content">
-
+<div #Container (resized)="onResize($event)" class="viewport-content w-100 h-100 flex-shrink-1">
+    <canvas #Canvas style="position: fixed;"></canvas>
```
marijnvanderhorst: Use CSS class instead

src/app/viewport/viewport.component.ts
Erik9519: Uses Event Emitter now

src/app/viewport/viewport.component.ts
Erik9519: Fixed and added short explanation.

src/app/viewport/viewport.component.html
Erik9519: Now uses CSS

src/app/viewport/viewport.component.scss
```diff
@@ -10,3 +10,7 @@
     bottom: 0;
     width:100%;
 }
+
+.viewportCanvas {
```
Dacaspex: For consistency, we use `kebab-case` for css selectors

Erik9519: Fixed mistake

PR #234, opened 2019-05-24
[do not merge] Disable headless mode for Firefox in Travis

Dacaspex:
**Merge only after we tested it with #232**

#### What have I done?
Issue #214 required a non-headless browser for Firefox in the Travis container, since a headless Firefox browser does not support webgl.


Dacaspex:
Will fix this in #232

Issue #100, opened 2019-05-15
Import container is too large on smaller screen


Vogeltak:
The import container is too large, causing the need of a scrollbar on the whole page. In such cases, it also becomes clear that the margin below the import button is insufficient to accommodate for pleasant viewing.

The height of the import container should be reduced on smaller screens (e.g. iPhone 5/SE).

See images below that illustrate the problem.

![image](https://user-images.githubusercontent.com/7063067/57776043-d0f5ba80-771e-11e9-8d2c-651bd4eb1c62.png)
![image](https://user-images.githubusercontent.com/7063067/57776397-aa844f00-771f-11e9-8ed6-933c79fcbc4a.png)

Chessnl:
I believe this problem is also relevant for desktop. As there is a jump in sizes when the import screen adds a scrollbar whereas the previous screen did not. See images below. 

![image](https://user-images.githubusercontent.com/34745596/57841972-e117a380-77cb-11e9-9935-6c4b6a610763.png)
![image](https://user-images.githubusercontent.com/34745596/57841859-a3b31600-77cb-11e9-8b3a-a2c609140591.png)

The images are not very clear so its better to see yourself. Go to the initial screen, scale it until it has no scrollbar but the import screen does and switch between them.

PR #251, opened 2019-05-27
#100 Import page sizing

ArtixOfVenion:
#### What have I done
Resolved issue with start page being too large, fixed sizing of start and import page and added scrollbar to manual import screen

#### Testing
Open the import page and try resizing it, should work on iphone 5. Try by hosting it locally and opening it on your phone. On manual import a scrollbar for the righthand side of the page should appear.

#### Addresses SRD section(s)
[No new functions or variables have been added]


src/app/import/import.component.scss
```diff
@@ -38,6 +44,10 @@
     overflow-y: auto;
 }
 
+.importManualOverflow {
```
cwognum: We already have a class for this!

PR #264, opened 2019-05-28
Feature/connect poi selector

TijnBertens:
#### What have I done
Connected the POISelector UI component to the POIService in a similar fashion as the SheetSelector is connected to the SheetService. POI names now synchronize between the POIService and the selector, and the same holds for the selected POI names. 

A dependency has been added from POIService to Path, such that the POIService can obtain a list of POIs. Furthermore, the PaperStateEngine calls an initialization function in the POIService.

#### Testing
Run unit tests.
Optionally you can manually add some mock POI names to the POI service and play around with the selector in the application. An easy way to do so, is by changing the following line in poi-service.ts:

```
    /**
     * List of the names of all POIs in the current path.
     */
    private poiNames: string[] = ['test1', 'test2' , 'test3' , 'test4'];
```

#### Addresses SRD section(s)
Functional requirements of:
POI selector
POI service
Sheet service


src/app/backend/poi-service/poi-service.ts
```diff
@@ -1,29 +1,73 @@
 /*
 Will be implemented after SheetService, to mimic behavior.
  */
```
nolmoonen: Remove this comment.

src/app/backend/poi-service/poi-service.ts
```diff
@@ -1,29 +1,73 @@
...
+        this.observablePoiNames.next(this.poiNames);
     }
 
     /**
      * Selects a poi name.
      * @return false if {@param poiName} already selected, true otherwise.
      */
     public selectPoiName(poiName: string): boolean {
-        if (this.selectedPoiNames.includes(poiName)) {
+        if (this.selectedPoiNames.includes(poiName) || !this.poiNames.includes(poiName)) {
```
nolmoonen: Good addition, but please also update method contract.

src/app/menubar/poi-selector/poi-selector.component.spec.ts
```diff
@@ -2,27 +2,216 @@ import { async, ComponentFixture, TestBed } from '@angular/core/testing';
...
+import { POI } from '../../backend/data-representation/poi';
+import { ControlPoint } from '../../backend/data-representation/control-point';
+import { Segment } from '../../backend/data-representation/segment';
+import { POIPosition } from '../../backend/data-representation/poiposition';
+import { POIService } from '../../backend/poi-service/poi-service';
+import { IllegalArgumentError } from '../../errors/illegal-argument-error';
+
+/**
+ * Checks whether two arrays have the same objects in them.
+ */
```
nolmoonen: Can you make it clear that the amount of duplicates may differ?

src/app/backend/poi-service/poi-service.ts
TijnBertens: fixed

src/app/backend/poi-service/poi-service.ts
TijnBertens: Fixed

src/app/menubar/poi-selector/poi-selector.component.spec.ts
TijnBertens: updated

Issue #263, opened 2019-05-28
POI selector resizes when it has a selection.


TijnBertens:
When the POI selector contains a selected element, it will resize/un-align itself slightly. See the following screenshots.

![image](https://user-images.githubusercontent.com/7274733/58466168-6b9bc380-8139-11e9-9f19-5032a8a7293e.png)

![image](https://user-images.githubusercontent.com/7274733/58466187-722a3b00-8139-11e9-9dc9-eb3b807e74c3.png)

cwognum:
Looked into this for a bit a couple of days ago. Has to do with a margin or padding that is added to the component that displays the selected item if I recall correctly

SquarePlayn:
This also applies to the dropdown of the table

PR #275, opened 2019-05-28
#125 #126 Feature/table

Chessnl:
#### What have I done
Create table with multiple subclasses for it that displays the data from sheetTimingStates and their POIMetrics. Columns and rows are configurable to choose which to display.

#### Testing 
To test on static data, you can paste the code from the following Pastebin to a MockSheetService and a MockPoiService to see the table be filled with data (as the current development does not have live data yet): https://pastebin.com/HUcRdYD4 . You can just put these in the `table.component.ts`. Make sure to change the injected dependencies of the Table component from SheetService to this MockSheetService and from PoiService to this MockPoiService

#### Known issues
- Table opening is slow
- Cells need to be trimmed
- Colors do not look nice yet
- Horizontal size of the table can grow too large (such that the graph does not fit on the side), but that is likely not caused by the table but rather the overlying styling.
- Dropdowns get a little less margin when no items are selected, see issue #263 
- When data keeps changing (e.g. when playing the visualization), the data may get wider / smaller columns all the time

#### Addresses SRD section(s)
3.1.22 Table


src/app/backend/poi-service/poi-service.ts
```diff
@@ -13,7 +13,7 @@ export class POIService {
     private poiNames: string[] = [];
 
     /**
-     * Contains the names of all the selected POIs.
+     * Contains the the names of all the selected POIs.
```
nolmoonen: I think something went wrong here

src/app/table/column-service/column.service.ts
```diff
@@ -0,0 +1,197 @@
...
+    observableAllColumns: BehaviorSubject<Column[]>;
+
+    constructor() {
+        let id = -1;
+        const staticColumns: Column[] = [
+            new Column(
+                ++id,
+                'Time',
+                () => 'Time',
+                undefined,
```
Dacaspex: Why not add a hover text to each column instead of passing `undefined`? Then you can add a check in the `Column` class that it should not be undefined and also you can remove the `if (hoverText === undefined)` check.

src/app/table/column-service/column.service.ts
```diff
@@ -0,0 +1,197 @@
...
+
+    /**
+     * observable of all columns regardless of dependency on POI's
+     * the columns array must have for all entries, located at index i that their id is i
+     * this variable is accessible as to verify its state in column.service.spec.ts
+     */
+    observableAllColumns: BehaviorSubject<Column[]>;
+
+    constructor() {
+        let id = -1;
```
Dacaspex: You already hardcoded the `Column` instances, then why bother keeping track of an `id` and increment this?

Having an id start at -1 can be confusing, if you do not agree with my comment above, then please change this to `let id = 0; ... id++, ...`

src/app/table/column-service/column.service.ts
```diff
@@ -0,0 +1,197 @@
...
+     */
+    public getColumnById(id: number): Column {
+        if (0 <= id && id < this.observableAllColumns.getValue().length) {
+            if (this.observableAllColumns.getValue()[id].getId() === id) {
+                return this.observableAllColumns.getValue()[id];
+            } else {
+                throw new IllegalStateError('observableAllColumns is not indexed by ids at index ' + id);
+            }
+        } else {
+            return undefined;
```
Dacaspex: I would throw an `IllegalArgumentError` instead. `null` and `undefined` are the root of all evil :)

src/app/table/entry/entry.ts
```diff
@@ -0,0 +1,64 @@
+import { IllegalArgumentError } from '../../errors/illegal-argument-error';
+
+/**
+ * An entry to be displayed in the table.component
+ * Entries can be either a number or a string. When both entries are of equivalent type, they can be compared
+ */
+export class Entry {
+
+    constructor(private value: string | number) {
```
Dacaspex: Please add this as a field as well `private value: string | number;`

Dacaspex: Please add this as a field as well `private value: string | number;`

src/app/table/entry/entry.ts
```diff
@@ -0,0 +1,64 @@
...
+
+        const value2 = entry.getValue();
+
+        if (this.value === undefined || value2 === undefined) {
+            return new Entry(undefined);
+        }
+
+        if (typeof this.value === 'string' && typeof value2 === 'string') {
+            if (this.value === value2) {
+                return new Entry('equal');
```
Dacaspex: This seems hacky :s

src/app/table/entry/entry.ts
```diff
@@ -0,0 +1,64 @@
...
+     * @throws IllegalArgumentError when one entry is a string and another is a number
+     * @throws IllegalArgumentError when the entry to be compared with is undefined
+     * @param entry to be compared with
+     */
+    public compare(entry: Entry): Entry {
+        if (entry === undefined) {
+            throw new IllegalArgumentError('entry is undefined');
+        }
+
+        const value2 = entry.getValue();
```
Dacaspex: `value2` sounds a bit dull, what about `otherValue`?

src/app/table/entry/entry.ts
```diff
@@ -0,0 +1,64 @@
...
+
+        if (this.value === undefined || value2 === undefined) {
+            return new Entry(undefined);
```
Dacaspex: Is this allowed to happen!? Why not throw an error instead?

src/app/table/entry/entry.ts
```diff
@@ -0,0 +1,64 @@
...
+        return this.value;
+    }
+
+    /**
+     * returns the value of this entry as a string
+     * returns '-' when its value is undefined
+     */
+    public toString(): string {
+        if (this.value === undefined) {
+            return '-';
```
Dacaspex: Ah, so if it is `undefined`, you map the value to `"-"`. Again, this sounds hacky to me as you are using `undefined` to mean something different. Instead, what about creating another class called `UndefinedEntry`? Or something similar

src/app/table/row-service/row.service.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * observable of all rows
+     * the rows array must have for all entries, located at index i that their id is i
+     * this variable is accessible as to verify its state in row.service.spec.ts
+     */
+    observableRows: BehaviorSubject<Row[]>;
+
+    constructor() {
+        let id = -1;
```
Dacaspex: Again the `id` thing, as I already talked about with the `ColumnService`

src/app/table/row-service/row.service.ts
```diff
@@ -0,0 +1,123 @@
...
+        const rows: Row[] = [
+            // First timing file loaded
+            new Row(
+                ++id,
+                'Timing file 1',
+                (sheetState: SheetState, numTimingFiles: number) => {
+                    if (numTimingFiles > 1) {
+                        return sheetState.getId() + '.1';
```
Dacaspex: Why the `.1`? Why only if there are more then 1 timing files? Please add a comment here explaining why this behaviour is in place

src/app/table/row-service/row.service.ts
```diff
@@ -0,0 +1,123 @@
...
+                (sheetState: SheetState) => {
+                    return sheetState.getId() + '.dif';
+                },
+                (sheetState: SheetState, poiName: string, column: Column) => {
+                    const entry0 = column.getValue(sheetState.getSheetTimingStates()[0], poiName);
+                    const entry1 = column.getValue(sheetState.getSheetTimingStates()[1], poiName);
+                    return entry0.compare(entry1);
+                },
+                (numTimingFiles: number, sheetState: SheetState, isSelectedRow: boolean) => {
+                    return !isSelectedRow && numTimingFiles > 1
```
Dacaspex: Complex proposition, maybe add some comments

src/app/table/row-service/row.service.ts
```diff
@@ -0,0 +1,123 @@
...
+     */
+    public getRowById(id: number): Row {
+        if (0 <= id && id < this.observableRows.getValue().length) {
+            if (this.observableRows.getValue()[id].getId() === id) {
+                return this.observableRows.getValue()[id];
+            } else {
+                throw new IllegalStateError('observableRows is not indexed by ids at index ' + id);
+            }
+        } else {
+            return undefined;
```
Dacaspex: Why not throw an `IllegalArgumentError`?

src/app/table/row/row.ts
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * the text which is displayed in the select-box for this given row
+     * needs to be public, as the ng-select reads from it
+     */
+    public readonly dropdownText: string;
+
+    /**
+     * defines how the id of this row is constructed
+     */
+    private readonly displayIdDefinition: (sheetState: SheetState, numTimingFiles: number) => string;
```
Dacaspex: The naming of these variables was giving me a hard time, what do you by `definition`? Why is this a function?

src/app/table/row/row.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * defines the value in the row for a given sheetTimingState, poiName and column
+     */
+    private readonly valueDefinition: (sheetState: SheetState, poiName: string, column: Column) => Entry;
+
+    /**
+     * defines whether this row can be displayed for given loaded numTimingFiles, sheetState and whether it is selected
+     */
+    private readonly displayDefinition:
```
Dacaspex: What does `isSelected` have to do with being able to be displayed or not? Being selected or not, whether it can be displayed should still hold right?

src/app/table/table.component.html
```diff
@@ -1,39 +1,113 @@
...
+                       (remove)="disableRow($event)"
+                       [searchable]=false
+                       [clearOnBackspace]=false
+                       [clearable]=false
+                       [closeOnSelect]=false
+                       [selectOnTab]=true>
+                <!-- This section allows you to specify what is displayed in the dropdown top bar -->
+                <!-- The most common use case is to construct the text of active items yourself -->
+                <!-- But here we just put the text in the field since we do not need the active items displayed -->
+                <!-- TODO Fix currently the below is not shown when nothing selected and only placeholder is shown -->
```
Dacaspex: TODO: If you are not going to fix this in this PR, then please create an issue and link it to the code :+1:

src/app/table/table.component.html
```diff
@@ -1,39 +1,113 @@
...
+            <thead>
+                <tr>
+                    <!-- ID column is always included -->
+                    <th scope="col" nowrap="nowrap">ID</th>
+                    <!-- All other columns are loaded in based on whether they are selected -->
+                    <!-- POI-independent columns -->
+                    <ng-container *ngFor="let column of columnService.getStaticColumns()">
+                        <th *ngIf="column.isEnabled()" scope="col" nowrap="nowrap">
+                        <span ngbTooltip={{column.getHoverText(undefined)}}>
+                           {{column.getHeaderText(undefined)}}
```
Dacaspex: Why do you pass `undefined` here? That doesn't look right

src/app/table/table.component.scss
```diff
@@ -0,0 +1,4 @@
+.table {
```
Dacaspex: `.table` is quite generic, perhaps make it more precise?

src/app/table/column-service/column.service.ts
Chessnl: Incrementing is done, such that when later on a new column is added, all preceding id's don't need to be adapted.

++id instead of id++ is done, as else it throws a warning, id is incremented but never used again (at the last increment).

src/app/table/column-service/column.service.ts
Chessnl: Changed to () => undefined, addapted constructer precondition and corresponding testcases.

src/app/table/column-service/column.service.ts
Chessnl: Should be resolved now.

src/app/table/entry/entry.ts
Chessnl: Should be resolved now.

Chessnl: Seems to be a duplicate comment? Resolved?

src/app/table/table.component.html
Chessnl: Made the poiName parameter an optional parameter. Should be resolved now.

src/app/backend/poi-service/poi-service.ts
Chessnl: Should be resolved now

src/app/table/row/row.ts
Chessnl: I would appreciate a suggestion as of how to resolve this. (can of course also explain why I chose definition as naming as well).

src/app/table/entry/entry.ts
Chessnl: Should be resolved.

src/app/table/entry/entry.ts
Chessnl: Let's say (for instance) we are comparing 2 timing files at time t. Paper p in timing file 1 is located at position 10, while paper p in timing file 2 is not in the system. Then for paper p in timing file 2, the position is undefined. Comparing this undefined position with an existing position does not make any sense, and hence nothing is displayed, while also no error should be thrown.

src/app/table/entry/entry.ts
Chessnl: The reason why we have a single Entry class instead of 3 specific Entry-classes extending from Entry depending on whether their value is undefined, number or string is that it is a significant amount of classes/lines that do not do anything. As instead of checking in a comparison whether you are comparing two numbers, you will still need to check whether you are comparing two numberEntries.

src/app/table/row-service/row.service.ts
Chessnl: See comment left at ColumnService.

src/app/table/row-service/row.service.ts
Chessnl: Because if we have a single timing file loaded we have for each row the ID's being only the paper ID. So 1, 2, 3, ...
However, when there are two timing files we append with which timing file a row corresponds. So 1.1, 1.2, 2.1, 2.2, 3.1, 3.2, ...
We add .1 only when there are more than 1 timing file loaded, as else we would have:
1.1, 2.1, 3.1, ... or 1, 1.2, 2, 2.2, 3, 3.2, ...

src/app/table/row-service/row.service.ts
Chessnl: Comment added.

src/app/table/row-service/row.service.ts
Chessnl: Should be resolved now

src/app/table/entry/entry.ts
Chessnl: Corresponds to comment about whether to use UndefinedEntry. Please forward any follow up comments to that thread.

src/app/table/table.component.scss
SquarePlayn: Changed to .margin-bot-0

src/app/table/table.component.html
SquarePlayn: Opened #295

src/app/table/row/row.ts
Chessnl: Added more elaborate comments for these variables. Please let me know if you think these are clear enough.

src/app/table/row/row.ts
Chessnl: Comment extended for elaboration. Please let me know if you think this is now clear enough.

src/app/table/row-service/row.service.ts
nolmoonen: But now you'd get `1, 1.1, 1.1, 2, 2.1, 2.1` with 3 timing files? Where you would want `1, 1.1, 1.2, 2, 2.1, 2.2`.

SquarePlayn: @nolmoonen Currently if there is only 1 timing file imported, you get "1". If there are more than 1 timing file imported, you get "1.1" (and "1.2" and "1.dif" for the other rows)

What do you mean with that you want `1, 1.1, 1.2, 2, 2.1, 2.2`? Having `1` ánd `1.1` is two rows with the same info, right?

SquarePlayn: Discussed with @nolmoonen in person as he cannot comment here anymore for some reason

src/app/table/column-service/column.service.ts
nolmoonen: I agree with both of you, but the current solution seems weird in the sense that it only avoids a warning? I would say go for `id = 0` and post-increment, or hard-code it (since this is already the case for `dropdownText`.

src/app/table/row/row.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+            false
+        );
+        expect(row.getValue(dummySheetState, dummyColumn, dummyPoiName).toString)
+            .toBe(valueDefinition(dummySheetState, dummyColumn, dummyPoiName).toString);
+        const dummyIsSelectedRow = false;
+        expect(row.canDisplay(dummyNumTimingFiles, dummySheetState, dummyIsSelectedRow))
+            .toBe(displayDefinition(dummyNumTimingFiles, dummySheetState, dummyIsSelectedRow));
+    });
+
+    it('should set and get enabled correctly', () => {
```
nolmoonen: `'should throw errors when passing undefined parameters in construction'`?

nolmoonen:
Also change the SRD section for SheetService, since methods have been renamed.

nolmoonen:
Add 
```
constructor(path: Path) {
        super(path);
}
```
to `MockPOIService`.

src/app/table/table.component.ts
```diff
@@ -7,7 +14,87 @@ import { Component, OnInit } from '@angular/core';
...
+        const column = this.columnService.getColumnById(columnId);
+        column.setEnabled(true);
+    }
+
+    /**
+     * Disable a column from a dropdown (remove) event
+     * @param event (remove) event generated by the column dropdown
+     */
+    disableColumn(event: any): void {
+        const columnId = event.value.id;
```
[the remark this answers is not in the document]
Chessnl: Added a comment in the code. But this is simply due to the format of events given by the HTML column being different for enable and disable events.

SquarePlayn:
> Change `MockPOIService` to:
> 
> ```
> @Injectable({
>     providedIn: 'root'
> })
> export class MockPOIService extends POIService {
> 
>     constructor(path: Path) {
>         super(path);
>     }
> 
>     public getSelectedPoiNames(): string[] {
>         return ['Soulja POI', 'It\'s dat POI, shit waddup'];
>     }
> }
> ```

PR info updated with new pastebin

src/app/table/row/row.spec.ts
Chessnl: Should be resolved now

src/app/table/column-service/column.service.ts
Chessnl: `id++` is now used. Last time instead of `id++`, `id` is used to circumvent the warning.

SquarePlayn:
Updated the compare function to class system with Entry, EmptyEntry, StringEntry, NumberEntry and EntryFactory.

src/app/table/table.component.scss
Dacaspex: There is a bootstrap class for that `mb-0` see [Bootstrap documentation](https://getbootstrap.com/docs/4.0/utilities/spacing/)

src/app/table/row/row.ts
Dacaspex: I would rename this to `map*` because it is a mapper function. Please update this for all files that use the `*Definition` syntax

src/app/table/row/row.ts
Dacaspex: As discussed, I would use `map*` for this

src/app/table/column-service/column.service.ts
Dacaspex: Why not simply do `() => ''`? Since you want to print an empty string, right?

src/app/table/row/row.ts
SquarePlayn: Updated function names to `map*`

src/app/table/row/row.ts
SquarePlayn: Updated function names to `map*`

src/app/table/table.component.scss
SquarePlayn: Updated to use `mb-0`

src/app/table/column-service/column.service.ts
SquarePlayn: If you provide undefined to a hover function it will not display a hover. We found that to be preferable to displaying an empty hover. Both do achieve this (probably, not 100% sure about empty string), but with undefined we find it to be clearer that it does not display any hover.

PR #276, opened 2019-05-28
Connect data to the graph

### LINKED ISSUE #189: [5] Connect graph with the data
Estimated: 8h

### LINKED ISSUE #283: [9] Test graph service

marijnvanderhorst:
#### What have I done
Fixes #189.
Fixes #283.
Fills the graph component with data calculated by the SheetStateEngine.

#### Testing
Import all files, commenting out the parse errors, and look at the graph.

#### Addresses SRD section(s)
Graph Service.


src/app/backend/graph-service/sheet-data/sheet-data.ts
```diff
@@ -0,0 +1,99 @@
...
+    public getPoints(graphMetric: GraphMetric): Point[] {
+        return this.metricPoints.get(graphMetric);
+    }
+}
+
+/**
+ * Class containing x,y pairs.
+ */
+export class Point {
+    private readonly _x: number;
```
nolmoonen: This does not conform to our naming scheme. Does this have to be like this?

nolmoonen:
I think it would be nice if you explain when you use `_variableName` instead of `variableName`, why this needs to be the case. Since it is now used interchangeably throughout classes.

src/app/backend/graph-service/sheet-data/sheet-data.ts
marijnvanderhorst: This is the way the TypeScript getters and setters roll. I'll add some comments

src/app/backend/graph-service/graph.service.ts
```diff
@@ -0,0 +1,251 @@
...
+        return array.reduce((acc, x) => acc.concat(map(x)), []);
+    }
+
+    /**
+     * Based on the max time and number of points, calculate the interval between two consecutive points.
+     *
+     * @param max The max time.
+     * @returns the calculated time interval.
+     */
+    calculateTimeInterval(max: number) {
```
nolmoonen: Return type?

src/app/backend/graph-service/graph.service.ts
```diff
@@ -0,0 +1,251 @@
...
+    /**
+     * First maps each element using a mapping function, then flattens the result into a new array.
+     * See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap
+     *
+     * @param array The array to apply the mapping to.
+     * @param map Function that produces an element of the new array, taking the current element being processed.
+     *
+     * @returns A new array with each element being the result of the {@code map} function and flattened to depth 1.
+     */
+    private flatMap(array: any[], map: (value: any) => any[]) {
```
nolmoonen: Return type?

src/app/backend/file-parser/parser.service.ts
```diff
@@ -51,6 +53,9 @@ export class ParserService {
                 this.timingParser.parse(timingData);
                 this.trackParser.parse(trackData);
 
+                // TODO find better place.
+                this.sheetStateEngine.initialize();
```
nolmoonen: a r c h i t e c t u r a l d e p t h

src/app/backend/graph-service/graph.service.spec.ts
```diff
@@ -0,0 +1,12 @@
+import { TestBed } from '@angular/core/testing';
```
nolmoonen: Maybe fix indentation of this file?

src/app/backend/file-parser/parser.service.ts
marijnvanderhorst: Will be addressed by @Dacaspex

src/app/backend/graph-service/graph.service.ts
```diff
@@ -0,0 +1,251 @@
...
+    private cloneSheetStates(sheetStates: SheetState[]): SheetState[] {
+        // Will eventually contain the result.
+        const clonedStates: SheetState[] = [];
+
+        for (const sheetState of sheetStates) {
+            // Will eventually contain the fresh timing states for this sheet state.
+            const newTimingStates: SheetTimingState[] = [];
+
+            // Cloned state to fill.
+            let newState: SheetState;
```
nolmoonen: Redundant variable?

src/app/backend/file-parser/parser.service.ts
Dacaspex: Will be addressed in #223

src/app/backend/graph-service/graph.service.ts
```diff
@@ -0,0 +1,251 @@
...
+
+    constructor(private time: TimeService,
+                private sheetService: SheetService,
+                private sheetStateEngine: SheetStateEngineService) {
+        // Initially no sheetData is present.
+        this.sheetData = [];
+        this.observableSelectedSheetData = new BehaviorSubject([]);
+
+        // Initially select the velocity metric in the dropdown.
+        this._selectedGraphMetric = GraphMetric.VELOCITY;
```
nolmoonen: `GraphMetric[0]`?

src/app/backend/graph-service/graph-metric.enum.ts
```diff
@@ -0,0 +1,7 @@
+export enum GraphMetric {
```
Dacaspex: I'm missing a comment here explaining what this enum is used for

src/app/backend/graph-service/graph.service.ts
marijnvanderhorst: Refactored.

src/app/backend/graph-service/sheet-data/sheet-data.ts
```diff
@@ -0,0 +1,99 @@
...
+     */
+    public getPoints(graphMetric: GraphMetric): Point[] {
+        return this.metricPoints.get(graphMetric);
+    }
+}
+
+/**
+ * Class containing x,y pairs.
+ */
+export class Point {
```
nolmoonen: `DataPoint`? `GraphPoint`? Also, how about `time, value` instead of `x, y`?

Dacaspex: Multiple exports in a file, only one class per file is allowed

src/app/backend/graph-service/graph.service.ts
marijnvanderhorst: Added

src/app/backend/graph-service/graph.service.ts
marijnvanderhorst: Fixed.

src/app/graph/graph.component.ts
```diff
@@ -1,7 +1,45 @@
...
-import {ResizedEvent} from 'angular-resize-event';
+import { Component, OnInit, ViewChild } from '@angular/core';
+import { ChartDataSets, ChartOptions, ChartType } from 'chart.js';
+import { BaseChartDirective, Color } from 'ng2-charts';
+import { ResizedEvent } from 'angular-resize-event';
+import { GraphService } from '../backend/graph-service/graph.service';
+import { SheetData } from '../backend/graph-service/sheet-data/sheet-data';
+import { GraphMetric } from '../backend/graph-service/graph-metric.enum';
+
+export const highlightColors = [
```
Dacaspex: What do the indices of this array correspond to? Can't these values be extracted in separate variables instead?

src/app/backend/graph-service/sheet-data/sheet-data.ts
marijnvanderhorst: Moved to own class and added comment.

src/app/backend/graph-service/graph-metric.enum.ts
marijnvanderhorst: Added comment.

src/app/backend/graph-service/sheet-data/sheet-data.ts
marijnvanderhorst: Fixed.

src/app/graph/graph.component.ts
marijnvanderhorst: Added comment.

src/app/backend/graph-service/graph.service.ts
```diff
@@ -0,0 +1,248 @@
...
+                    sheetState.getId(),
+                    timingState.getTimingId(),
+                    this.sheetService.getSheetName(timingState)
+                ));
+            }
+        }
+    }
+
+    /**
+     * Given a list of sheet states, goes over all sheet timing states and adds a new data point to {@code sheetData},
```
Dacaspex: It is not given a list of `sheetStates` but only one sheet state

src/app/backend/graph-service/graph.service.ts
```diff
@@ -0,0 +1,248 @@
...
+        if (this.time.getMaxTime() === 0 || this.timeInterval === 0) {
+            // There is nothing to compute.
+            return;
+        }
+
+        // Create a new sheet state for each sheet state.
+        const statesForComputation = this.cloneSheetStates(sheetStates);
+
+        // Initialize the sheetData array.
+        this.initializeSheetData(sheetStates);
```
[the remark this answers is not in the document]
marijnvanderhorst: I don't understand your reasoning here. `cloneSheetStates` does indeed clone the sheet states, returning a new array. You say `initializeSheetData` populates the `sheetData` as a side-affect, whereas it's the one and only thing it does. In fact it's in the name of the function, initializing the `sheetData` array. Thus, I wouldn't say it does it as a side-affect. The solution you propose, initializing the `sheetData` during cloning, would result in adding a side-affect to the clone function.

> Why do we need to initialize something with the sheet states, if we already cloned them? Shouldn't initializing only happen at the first time?"

The sheet states are not initialized here, the `sheetData` is. These are two very separate things, hence why they are separated into two separate functions.

The reasoning about `sheetData` is that it is a tuple of a sheet and a data set, resulting from the one-to-one relation. If you can think of a better name, please let me know.

src/app/backend/graph-service/graph.service.ts
marijnvanderhorst: Fixed.

src/app/backend/graph-service/graph.service.ts
```diff
@@ -0,0 +1,248 @@
...
+        return max / GraphService.NR_OF_POINTS;
+    }
+
+    /**
+     * Clones the SheetState objects into new ones, where each contained timing state is a fresh new one.
+     *
+     * @param sheetStates The sheet states to clone.
+     * @returns the cloned sheet states.
+     */
+    private cloneSheetStates(sheetStates: SheetState[]): SheetState[] {
```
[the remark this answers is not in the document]
marijnvanderhorst: This is really a function that belongs to the graph service singleton. Given this is already a singleton, what is your reason behind making this function static, rather than keeping a nice and beautiful instance method, following the good old OO-principles :)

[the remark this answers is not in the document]
marijnvanderhorst: Now say, later on, someone wants to extend the graph service, and use the dependency injection system provided by Angular to inject the sub-class instead. Then there is no possibility at all to override this method, is that really what you want?

src/app/backend/graph-service/graph.service.ts
```diff
@@ -0,0 +1,248 @@
...
+        return array.reduce((acc, x) => acc.concat(map(x)), []);
+    }
+
+    /**
+     * Based on the max time and number of points, calculate the interval between two consecutive points.
+     *
+     * @param max The max time.
+     * @returns the calculated time interval.
+     */
+    calculateTimeInterval(max: number): number {
```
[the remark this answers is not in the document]
marijnvanderhorst: Now say, later on, someone wants to extend the graph service, and use the dependency injection system provided by Angular to inject the sub-class instead. Then there is no possibility at all to override this method, is that really what you want?

src/app/backend/graph-service/graph.service.ts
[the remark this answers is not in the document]
marijnvanderhorst: For consistency and testability, I will refactor the `initializeSheetData` to create the array, and set it to the instance variable inside the `compute` function. I'll also rename the function as a consequence.

PR #281, opened 2019-05-29
#280 Added animation stats on mobile

cwognum:
#### What have I done
On mobile, I displayed the animation stats. See below: 
image

#### Testing
Serve the application and verify that the stats are correctly displayed on both mobile and desktop devices.


cwognum:
Sloppy fix... Might be better to create a new branch and a new PR

PR #285, opened 2019-05-29
Fix #282 - Disabled time slider

### LINKED ISSUE #282: Animation slider cannot be slided
Using the slider you should be able to click anywhere on the slider to skip to that timepoint. You should also be able to drag the timepoint. Both do not work anymore.

SquarePlayn:
#### What have I done
Fix that the time slider was enabled by default, disabled when a file was imported

#### Testing
Without importing files, note that you cannot slide the time slider. Import files (NB: Currently need to comment out the error throwing in the path parser for this) and note that you can now slide the time slider


marijnvanderhorst:
Did you #gitBlame the one responsible for this?

Issue #222, opened 2019-05-23
Change the implementation of FileReader to something that can work with fake async


Dacaspex:
The current implementation of the parse service uses the `FileReader`. Unfortunately, that implementation does not work with fake async, which is required in order to properly test the parse function. 

Therefore, we need an implementation of a file reader that works with fake async / macrotasks / promises.

Dacaspex:
Related Github issue https://github.com/angular/zone.js/issues/1120

Dacaspex:
> @Dacaspex one workaround: use new Response(file).arrayBuffer() instead of FileReader, with promise support

Dacaspex:
The `Response().then()` implementation does not seem to work as well, as the promises are not properly flushed in the `fakeasync` environment

cwognum:
@Dacaspex This is fixed, right? Can it be closed?

PR #289, opened 2019-05-29
#223 import second timing file

### LINKED ISSUE #223: Import second timing file

Dacaspex:
fixes #223 
#### What have I done
- Extended the `ParserService` and `ImportComponent` to be able to import a second timing file.
- Removed the `FileService` as it was giving major headaches on Firefox for some reason and after some inspection, it did not seem necessary to have. 

#### Testing
Import a second fiming file and see that an error is given (in the console) when the parsing fails, or the user is navigated to the workspace view when parsing succeeds

#### Addresses SRD section(s)
- SRD section about `ParserService`: One function has changed completely

**Note:** the low code coverage is due to #222


cwognum:
I just love how @SquarePlayn changed the title once already and there still is a typo in there!

src/app/backend/file-parser/parser.service.ts
```diff
@@ -66,22 +107,30 @@ export class ParserService {
      * Internal function to create an array of promises that read files
      * @param files Array of files to read
      */
-    private createFileReadPromises(files: File[]): Promise<string>[] {
+    private createFileReadPromiseList(files: File[]): Promise<string>[] {
```
TijnBertens: Since the code in ```parse()``` and ```parseWithExtraTimingFile``` depend explicitly on the order of elements returned by this function:
```
        // The order of "contents" will have the same order as defined
        // in this array
```
It may be wise to add this to the contract of this function.

src/app/import/import.component.ts
```diff
@@ -114,13 +109,7 @@ export class ImportComponent implements OnInit {
...
-        this.pathFile = this.fileStorage.getFiles().pathFile;
-        this.timingFileOne = this.fileStorage.getFiles().timingFileOne;
-        this.trackFile = this.fileStorage.getFiles().trackFile;
-        this.modelFile = this.fileStorage.getFiles().modelFile;
-        this.timingFileTwo = this.fileStorage.getFiles().timingFileTwo;
-        this.errorFile = this.fileStorage.getFiles().errorFile;
+        // Intentionally left blank
     }
 
     /**
```
TijnBertens: Not part of this PR, but the contracts of the next three functions seem to have been copied without being edited. Can you fix this in this PR?

src/app/import/import.component.ts
```diff
@@ -266,31 +255,42 @@ export class ImportComponent implements OnInit {
      */
     importManually() {
         if (!this.canImport()) {
-            // 'Cannot import manually; not all necessary files are present'
+            // Cannot import manually; not all necessary files are present
```
nolmoonen: Maybe it is better to put this in the error message?

src/app/import/import.component.ts
```diff
@@ -266,31 +255,42 @@ export class ImportComponent implements OnInit {
...
+
+        // Handler when parsing fails
+        const onFailure = (error: Error) => {
+            console.error('An error occurred during parsing');
+            console.error(error);
+        };
+
+        // Handler when parsing succeeds
+        const onSuccess = () => {
+            this.router.navigate(['/']);
```
TijnBertens: Can you add a comment like:
```// Navigate to home if parsing is successful```
Maybe my fault, but I didn't know what this constant ```'/'``` meant.

File: src/app/import/import.component.ts [hunk 5]
```diff
@@ -266,31 +255,42 @@ export class ImportComponent implements OnInit {
...
-            },
-            (error) => {
-                console.log('error');
-                console.log(error);
-            }
-        );
+        const pathFile = this.pathFile.item(0);
+        const timingFileOne = this.timingFileOne.item(0);
+        const trackFile = this.trackFile.item(0);
+        const timingFileTwo = this.timingFileTwo !== undefined
```
nolmoonen: If you break up the ternary operator, you might as well just make this an if-statement.

src/app/import/import.component.ts
```diff
@@ -266,31 +255,42 @@ export class ImportComponent implements OnInit {
...
+            ? this.timingFileTwo.item(0)
+            : undefined;
+
+        // Handler when parsing fails
+        const onFailure = (error: Error) => {
+            console.error('An error occurred during parsing');
```
nolmoonen: I think this is redundant. The message of the thrown error should be sufficient.

src/app/import/import.component.ts
nolmoonen: Which is more common and understandable to more people.

src/app/backend/file-parser/parser.service.ts
nolmoonen: What does this breaking up add? This will never be reused and makes the file longer and less clear.

src/app/backend/file-parser/parser.service.ts
```diff
@@ -54,10 +54,51 @@ export class ParserService {
...
+     * Parses the path, timing and track file asynchronously into the data representation.
+     *
+     * @param pathFile Path file
+     * @param timingFile Timing file
+     * @param extraTimingFile Extra timing file
+     * @param trackFile Track file
+     * @param onSuccess Callback upon success
+     * @param onFailure Callback upon failure, receives the error why it fails
+     */
+    parseWithExtraTimingFile(
```
TijnBertens: This function is not called in any test case, is this because of the bug mentioned in the spec:
```
        // Cannot test functionality properly because of a bug with FileReader
        // See https://github.com/angular/zone.js/issues/1120
        // The FileReader does not work properly with Angular fake async time
```

If so, then for the sake of consistency/coverage, since there is a test case for ```parse()```, shouldn't there also be one for ```parseWithExtraTimingFile()```. Even if the functionality can not properly be tested, the aforementioned comment would still apply and explain why exactly this function is not tested.

src/app/import/import.component.ts
Dacaspex: The way JavaScript error tracing works, I think it is beneficial to add these kind of messages. It becomes immediately clear where the error happened (as this might be an error that is thrown because of malformed user input)

src/app/backend/file-parser/parser.service.ts
Dacaspex: Maintainability and SRP mostly

src/app/import/import.component.ts
Dacaspex: Fixed

src/app/import/import.component.ts
Dacaspex: Fixed, made an if-statement

src/app/backend/file-parser/parser.service.ts
Dacaspex: Merged the two functions

Dacaspex: Fixed

src/app/import/import.component.ts
Dacaspex: Fixed

src/app/import/import.component.ts
Dacaspex: Fixed

src/app/backend/file-parser/parser.service.ts
Dacaspex: Fixed, added test case

Issue #303, opened 2019-06-03
Refactor codebase to group logical components into modules


nolmoonen:
As specified in [Angular docs](https://angular.io/guide/ngmodules). This will improve maintainability and help write documentation.

PR #312, opened 2019-06-04
Enhancement/path parser

Erik9519:
#### What have I done
Transformed certain errors into warnings instead.
POIs present in SegmentApperance but not their respective Segment are now fixed if they are either the first or last control point in the SegmentAppearance.

#### Testing
Unit testing

#### Addresses SRD section(s)
None


nolmoonen:
Please update from development see #303 .

src/app/core/services/parser/path-parser.service.ts
```diff
@@ -280,19 +280,35 @@ export class PathParserService {
...
                 }
             }
             // ERROR CHECK - Ensure that all POI's found in this Segment are specified in the respective
             //               SegmentAppearance
-            if (poisInSegment.length > 0) {
-                throw new ParserError(
-                    'Not all POI\'s found in the SegmentAppearance of Id ' + segId + ' have been found in the ' +
-                    'respective Segment.'
-                );
+            while (poisInSegment.length > 0) {
```
TijnBertens: As I understand it. At this point in the code, all points from the segment appearance are already removed from `poisInSegment`. So the only ones that are left are those that are in the segment but not in the segment appearance. Then you check if those are the first or last in the segment, in which case you add them to the path. If they are not, you throw a warning.

Why exactly is there no warning in the cases where it is the first or last POI? And why is it necessary to add them anyway? Could you explain this also in comments in the code, along with what I typed in the first two sentences of the previous paragraph?

PR #315, opened 2019-06-04
Fixes issue #308 : selects best placement for tooltip automatically instead of always placing tooltip at bottom.

Astrejoe:
#### What have I done
Fixed issue #308. It now places the tooltip for the speed down button automatically in the best place instead of forcing it underneath the button.

#### Testing
However over the speed down button. Observe that it doesn't cause scrollbar to appear.

#### Addresses SRD section(s)
/


nolmoonen:
Please update from development see #303 .

Astrejoe:
> Please update from development see #303 .

done

PR #367, opened 2019-06-06
#351 fixed async loading of model

### LINKED ISSUE #351: Properly load model before navigating to workspace
Right now one would need to go to the help view, before one can see the model. 

The model parsing happens async, but the parser service does not wait for this to finish. Required changes:
- Wait in the parser service for the model parsing to finish, then execute the success callback

Dacaspex:
fixes #351 
#### What have I done
The importing of the model happens async but the `ParserService` did not wait for the model loading to be complete. Therefore, the application already navigated to the workspace before the model was loaded.

Fixed that the `ParserService` waits for the model loading to complete

#### Testing
1. Import a 3D model
2. Navigate to the workspace
3. See that the 3D model is immediately there

#### Addresses SRD section(s)
Probably none, as I did not change any implementation, only the way we handle callbacks/promises which are implementation details


Dacaspex:
Low code coverage due to #222

PR #392, opened 2019-06-06
#324 Table no longer grows past the col size

### LINKED ISSUE #324: [Table] Set fixed / constant width of tale cells so they do not keep growing/shrinking
Partially related to #322

cwognum:
#### What have I done
Made sure the table does not expand beyond its Bootstrap column size. Fixes #324 

#### Testing
Check if it works in both Firefox and Chrome by resizing the screen and opening all three panels and making sure the panels each are of equal width


marijnvanderhorst:
Quite a lot of changes for a single PR, but nice fix

PR #393, opened 2019-06-06
Fixes #322 Trimming table cells

### LINKED ISSUE #322: [Table] Trim/formal cell values

SquarePlayn:
#### What have I done
Trimmed the values of number and string fields to constant lengths.

#### Testing
Automated testing.
Alternatively, load in data from the USB stick using the following change:
- Inject sheetStateEngine into the dispatch component and add to the update() function the line `this.sheetStateEngine.updateSheetService();`.
- Then in the sheetStateEngine's updateSheetService() function, change the content of the _if_ statement to just return (so don't throw an error).

#### Addresses SRD section(s)
Entry


SquarePlayn:
> The text color in the cells is currently black, which is not really very readable.

While I would argue that this regards #323, I have changed the color of the data to white.

src/app/table/entry/string-entry.ts
```diff
@@ -44,7 +49,12 @@ export class StringEntry implements Entry {
      * returns the value of this Entry to be displayed
      */
     public getDisplayValue(): string {
-        return this.value;
+        let padded = this.value;
+        for (let i = 0; i < StringEntry.STRING_LENGTH; i++) {
```
Dacaspex: As said in the earlier pull request, could you add a comment here explaining what you are doing?

SquarePlayn: Added comments

PR #403, opened 2019-06-11
Fix/column testcases utp

Chessnl:
#### What have I done
I have written a section for the UTP, which indicated a 'need' for more test cases then were currently in place. I updated the test cases ran for the Column and Columnservice in accordance to the UTP.

#### Testing
-

#### Addresses SRD section(s)
-


src/app/table/column-service/column.service.ts
```diff
@@ -267,7 +267,8 @@ export class ColumnService {
      */
     getColumnById(id: number): Column {
         if (0 <= id && id < this.observableAllColumns.getValue().length) {
-            if (this.observableAllColumns.getValue()[id].getId() === id) {
+            if (this.observableAllColumns.getValue()[id] !== undefined &&
```
Dacaspex: Is that check necessary? You already check if the `id` is in the range `[0, length]`?

Chessnl: Discussed with Casper. Resolved.

PR #406, opened 2019-06-11
Fix #404 by filtering out POIs w/o coords for the names list

### LINKED ISSUE #404: Add warning for POIs that do not have coordinates and make them not selectable
Estimate: 1h

Vogeltak:
#### What have I done
POI instances that do not have a position on the route should not be selectable. Thus we filter out those respective instances when initializing POI-service and only add the names of the POI instances that have a position on the route.

#### Testing
Import files and check that POI instances that are not on the route do not appear in the poi-selector. Run tests and verify that all pass.

#### Addresses SRD section(s)
Updated descriptions in 3.1.23.


src/app/core/services/poi-service.ts
```diff
@@ -39,9 +39,18 @@ export class POIService {
         this.clearPoiNames();
         this.clearSelectedPoiNames();
 
-        for (const poi of this.path.getPOIs()) {
-            this.addPoiName(poi.getName());
+        // Retrieve the POI instances to filter.
+        let pois = this.path.getPOIs();
+
+        // Warn the user if there exist POI instances that have no position on the route.
+        if (pois.some(poi => poi.getPositionsOnRoute().length === 0)) {
```
Dacaspex: Perhaps we could log all POIs that are skipped? Something like
```
pois
    .filter(poi => poi.getPositionOnRoute().length === 0)
    .forEach(poi => console.warn(poi));
```

Vogeltak: I agree it would be useful, but that results in 64 warning lines in the console.

Vogeltak: I have rewritten it such that the console warning now include all names of the POI instances that are filtered out as a list of strings separated by comma's. See a1f7180504cd93a48fe5109fde15f7d66641c313.

PR #412, opened 2019-06-11
Fix/row use centralized name

SquarePlayn:
#### What have I done
There is a centralized function in the sheetState to get the display ID. I changed the table to use that (and added some functionality around it to keep things in the same place).

#### Testing
Automated testing. Else load 1 and/or 2 timing files and check first column

#### Addresses SRD section(s)
Table sections. SheetService. Class Diagram


src/app/core/services/sheet.service.ts
```diff
@@ -193,6 +193,11 @@ export class SheetService {
         return name + '.' + timingId;
     }
 
+    getDifName(sheetState: SheetState): string {
```
Dacaspex: Doc comment?

src/app/table/entry/string-entry.ts
```diff
@@ -44,7 +49,12 @@ export class StringEntry implements Entry {
      * returns the value of this Entry to be displayed
      */
     public getDisplayValue(): string {
-        return this.value;
+        let padded = this.value;
```
Dacaspex: What is happening here precisely? I think adding comment helps

SquarePlayn: Addressed in #393

src/app/core/services/sheet.service.ts
SquarePlayn: Added doc comment

PR #423, opened 2019-06-13
Table colors 

### LINKED ISSUE #323: [Table] Theming / set proper colors

SquarePlayn:
#### What have I done
Added better colors to the table
Made the spacing of the table use `rem` instead of `px`
Fixes #323 

#### Testing
Load one and/or two timing files into the page.
Verify that all colors look nice.

#### Addresses SRD section(s)
None


SquarePlayn:
> screenshot_colors
> 
> I would personally make the color of the top bar with labels, the unselected rows and the background color (indicated by the blue boxes), the same color as the background of the graph, which is the same as the background color of the configuration (indicated by the purple box).

Good point. Updated it to use the centralized bg-dark color. 
what it looks like now

PR #427, opened 2019-06-13
Fix graph crashing in some cases

### LINKED ISSUE #384: Graph crashes when it is not opened before importing data
Steps to reproduce:
1. Open Exhibit
2. Do not open the graph panel
3. Import files
4. Try opening the graph panel
    - Observe that the application crashes

1. Open Exhibit
2. Open the graph panel
3. Import files
4. Open the graph panel
    - Observe that the application does not crash

marijnvanderhorst:
#### What have I done
Fixes #384.
The graphservice will not construct whenever an error occurred in the sheet state engine. 
To fix this, I make sure the sheet state engine does not use a position outside the route, which was the cause of the error.

#### Testing
* Open Exhibit
* Do not open the graph component
* Import files
* Observe the graph not crashing.

#### Addresses SRD section(s)
No changes required.


src/app/core/models/path/route.ts
```diff
@@ -31,17 +31,17 @@ export class Route {
             throw new IllegalStateError('segment positions not set');
         }
 
+        if (this.length === undefined) {
```
TijnBertens: Please update js doc

marijnvanderhorst: Done

marijnvanderhorst:
SRD has been updated.

PR #452, opened 2019-06-14
#435 and x-grid now shown by default

Erik9519:
#### What have I done
#435 addressed and x-grid now shown by default

#### Testing
Try to set the camera using the scene settings

#### Addresses SRD section(s)
ConfigurationService


src/app/workspace/viewport/viewport.component.ts
```diff
@@ -258,6 +258,9 @@ export class ViewportComponent implements AfterViewInit, OnDestroy {
             this.groundPlane.visible = this.configuration.getSceneShowGround();
             this.ambientLight.visible = this.configuration.getSceneShowAmbientLight();
             this.directionalLight.visible = this.configuration.getSceneShowDirectionalLight();
+            if (this.configuration.getSceneUsePerspectiveCamera() !== this.cameraProvider.getCamera().isPerspective()) {
```
Chessnl: To me, it is not immediately clear what this if-statement checks. Add a comment for elaboration. Also, add a test case covering this case.

src/app/configuration/configuration-service/configuration.service.spec.ts
```diff
@@ -61,4 +62,11 @@ describe('ConfigurationService', () => {
             );
         }
     });
+
+    it('should be able to set camera view', () => {
+        const service: ConfigurationService = TestBed.get(ConfigurationService);
+        expect(service.getSceneUsePerspectiveCamera()).toBeTruthy();
+        service.setSceneUsePerspectiveCamera(false);
+        expect(service.getSceneUsePerspectiveCamera()).toBeFalsy();
+    });
```
Chessnl: I would extend the test case, as it only checks whether it can be toggled off, not on.

```suggestion
        service.setSceneUsePerspectiveCamera(true);
        expect(service.getSceneUsePerspectiveCamera()).toBeTruthy();
    });
```

src/app/workspace/viewport/viewport.component.ts
Erik9519: Addressed

src/app/configuration/configuration-service/configuration.service.spec.ts
Erik9519: Addressed

PR #454, opened 2019-06-14
Make sure all TimingEntries are shown in the graph

### LINKED ISSUE #437: More dots in the graph for each timing entry

marijnvanderhorst:
#### What have I done
Fixes #437.
After the normal points have been calculated for the graph, also add a point for each timing entry.

#### Testing
Load files, look at graph.

#### Addresses SRD section(s)
/


src/app/graph/graph-service/graph.service.ts
```diff
@@ -235,6 +283,12 @@ export class GraphService {
             }
         }
 
+        this.addTimingEntriesToSheetData();
```
Chessnl: I would suggest adding a comment here.

marijnvanderhorst: Added

PR #459, opened 2019-06-17
Feature - Clearing of data representation and scene

Dacaspex:
#### What have I done
Implemented the clear button to actually clear the data representation and scene.
- Added `ClearService` that has dependencies to all components that need to be cleared
- Added `clear()` or `reset()` functions to required components.

#### Testing
- Import files, run the simulation for a while, select some sheets and POIs. Then clear and observe that everything works as expected.
- Import files with multiple timing files. Do the same as above.
- Change the camera / scene settings, do the same as above. 

#### Addresses SRD section(s)
- `ConfigurationService`
- `PrinterModel`
- `CameraProviderService`
- Added `ClearService`
- `TimeService`


src/app/core/services/time.service.ts
```diff
@@ -229,4 +219,21 @@ export class TimeService {
...
+
+    /**
+     * Resets the time service to its default values
+     */
+    reset() {
+        // Assign default initial values to class variables
+        this.currentTime = 0;
+        this.playbackSpeed = 1.0;
+        this.paused = true;
+        // Initialize maximum time to equal 0
```
Chessnl: I think this comment does not add much information. Besides that, in clear I would no longer call it initialize but rather set, as it can already be initialized. Regardless, I would omit this comment.

src/app/core/services/time.service.ts
```diff
@@ -229,4 +219,21 @@ export class TimeService {
...
+        this.maxTime = 0;
+        // Initialize minimum time to equal 0
+        this.minTime = 0;
+
+        // There is no previous system time if there have been no updates yet
```
Chessnl: I also think (as above) that there is little value to this comment and is no longer precise, seeing that it is not done once in the constructor.

src/app/import/import.component.ts
```diff
@@ -299,5 +304,13 @@ export class ImportComponent implements OnInit {
             );
         }
     }
+
+    /**
+     * Clears the files and data representation
+     */
+    clear() {
```
Chessnl: I would consider merging this function with `clearFiles()`.

Dacaspex: I would leave them separate as they concern different things (local clearing and global clearing)

src/app/configuration/configuration-service/configuration.service.ts
```diff
@@ -178,4 +173,17 @@ export class ConfigurationService {
     setSceneUsePerspectiveCamera(value: boolean): void {
         this.usePerspectiveCamera.value = value;
     }
+
+    /**
+     * Sets the default configuration settings
+     */
+    clear() {
+        this.configuration = new Array(3);
+        // Scene Configuration
```
Chessnl: These comments don't have any real meaning.

src/app/core/services/camera-provider.service.ts
```diff
@@ -25,4 +18,24 @@ export class CameraProviderService {
     getCamera(): Camera {
         return this.camera;
     }
+
+    /**
+     * Sets the camera to the default settings
+     */
+    reset(): void {
+        this.camera = new Camera(
```
Chessnl: The commenting as was done initially (behind each variable) was nicer to read.

src/app/core/services/clear.service.ts
```diff
@@ -0,0 +1,73 @@
...
+        this.route.setLength(undefined);
+        this.route.setSegmentPositions(undefined);
+
+        // Clear the configuration to the default settings
+        this.configurationService.clear();
+
+        // Reset the camera to its default settings
+        this.cameraProviderService.reset();
+
+        // Clear sheet service
```
Chessnl:
```suggestion
        // Clear the sheet service
```

src/app/core/services/clear.service.ts
```diff
@@ -0,0 +1,73 @@
...
+        this.sheetService.clearSelectedSheetStates();
+        this.sheetService.clearSheetStates();
+        this.sheetService.setNumTimings(0);
+
+        // Clear POI service
```
Chessnl:
```suggestion
        // Clear the POI service
```

src/app/configuration/configuration-service/configuration.service.ts
Dacaspex: Code was moved from a different method. I think they do add value

src/app/core/services/camera-provider.service.ts
Dacaspex: Because IntelliJ puts placeholders before each parameter, the comments were not aligned which was very annoying to read

src/app/import/import.component.ts
```diff
@@ -314,5 +319,13 @@ export class ImportComponent implements OnInit {
...
```
Chessnl: This function also is currently not tested, so add a test case for this.

Dacaspex: Fxied

PR #460, opened 2019-06-17
#456 Set time to paused when leaving workspace

Chessnl:
I am not sure whether this change is appropriate as a dependency between workspace and TimeService is made. I would like feedback on whether there are better alternatives.

#### What have I done
When the workspace component is destroyed, time is now paused.

#### Testing
Open the workspace. Set time to play. Leave the workspace (open help / open import / on mobile open configuration) and return. Time should be paused.

#### Addresses SRD section(s)
 -


Chessnl:
Discussed with Erik. He considers this a proper solution. Regardless, I still wish to have an additional review.

PR #461, opened 2019-06-17
Changed import menu to specify real and expected timing files

TijnBertens:
#### What have I done
Changed the import menu to indicate what imports are for the real and expected timing files. This is still a part of fixing issue #434.

#### Testing
Automated testing.
Check that the import menu shows a real and expected import and when either or both are selected, it is possible to import.

#### Addresses SRD section(s)
Import


src/app/import/import.component.ts
```diff
@@ -172,16 +166,16 @@ export class ImportComponent implements OnInit {
...
+        this.timingFileReal = files;
     }
 
     /**
      * Sets the second timing file.
      * @param files array containing the one file that needs to be set
      */
-    inputTimingTwo(files: FileList) {
-        this.timingFileTwo = files;
+    inputTimingExpected(files: FileList) {
```
Erik9519: Missing return type

src/app/import/import.component.ts
```diff
@@ -172,16 +166,16 @@ export class ImportComponent implements OnInit {
      * Sets the first timing file.
      * @param files array containing the one file that needs to be set
      */
-    inputTimingOne(files: FileList) {
-        this.timingFileOne = files;
+    inputTimingReal(files: FileList) {
```
Erik9519: Missing return type

src/app/import/import.component.ts
```diff
@@ -237,18 +231,17 @@ export class ImportComponent implements OnInit {
      */
     clearFiles() {
```
Erik9519: Missing return type

src/app/import/import.component.ts
TijnBertens: Fixed

src/app/import/import.component.ts
TijnBertens: Fixed

src/app/import/import.component.ts
TijnBertens: Fixed

Issue #453, opened 2019-06-14
Increase code coverage


Chessnl:
When checking others pull requests, I notice that there are here and there some uncovered lines, which are 'relatively easy' to cover (e.g. `keyEvent()` in `viewport.component.ts`). If time permits, I think it would be valuable to check for each non-covered line, whether it can be covered via some test cases.

PR #463, opened 2019-06-17
#444 Enhancement/Sheet selection via mouse click, visualization of sheets difference

Erik9519:
#### What have I done
Sheet selection via mouse click. Visualization of difference between sheets when mouse hover #444.

#### Testing
Try to click on sheets to select and deselect. Hover mouse on a sheet when 2 timing files are loaded and notice the text.

#### Addresses SRD section(s)
Viewport


src/app/workspace/viewport/viewport.component.ts
```diff
@@ -357,6 +435,36 @@ export class ViewportComponent implements AfterViewInit, OnDestroy {
...
+                                this.raycaster.intersectObject(sheetModel.getModel().children[0]);
+                            if (intersection[0] !== undefined) {
+                                let txt: string;
+                                // Check whether or not the sheet is alive in all SheetTimingStates
+                                if (sheetState.getSheetTimingStates()[0].getIsAlive() &&
+                                    sheetState.getSheetTimingStates()[1].getIsAlive()) {
+                                    txt = Math.abs(sheetState.getSheetTimingStates()[0].getPosition() -
+                                        sheetState.getSheetTimingStates()[1].getPosition()).toPrecision(4) + ' mm';
+                                } else {
+                                    txt = '- mm';
```
cwognum: We would change this to "Nothing to compare" or "Only one sheet" or something similar

Erik9519: Changed and fixed everything

src/app/workspace/viewport/utils/ray-cast.ts
```diff
@@ -0,0 +1,69 @@
...
+                        sheetState,
+                        timingID: sheetTimingState.getTimingId()
+                    });
+                }
+            }
+        }
+        return result;
+    }
+
+    checkForHitOnSheet(
```
Dacaspex: Doc comment

src/app/workspace/viewport/utils/ray-cast.ts
```diff
@@ -0,0 +1,69 @@
...
+     */
+    buildListOfSheetsThatCanBeHit(
+        sheetStates: SheetState[],
+        sheetModels: Map<number, Map<number, SheetModel>>
+    ): {sheet: SheetModel, sheetState: SheetState, timingID: number}[] {
+        const result: {sheet: SheetModel, sheetState: SheetState, timingID: number}[] = [];
+        // Build array of sheet models that are alive
+        for (const sheetState of sheetStates) {
+            for (const sheetTimingState of sheetState.getSheetTimingStates()) {
+                if (sheetTimingState.getIsAlive()) {
```
Dacaspex: No check whether the sheet is visible or not

src/app/workspace/viewport/utils/ray-cast.ts
```diff
@@ -0,0 +1,69 @@
...
+        return result;
+    }
+
+    checkForHitOnSheet(
+        sheets: {sheet: SheetModel, sheetState: SheetState, timingID: number}[]
+    ): SheetState {
+        // Build List of Object3D that can be hit
+        const object3D: THREE.Object3D[] = [];
+        for (const sheet of sheets) {
+            object3D.push(sheet.sheet.getModel().children[0]);
```
Dacaspex: Comment why using index `0`

src/app/workspace/viewport/utils/ray-cast.ts
Erik9519: Added

src/app/workspace/viewport/utils/ray-cast.ts
Erik9519: Checked

src/app/workspace/viewport/utils/ray-cast.ts
Erik9519: Added

src/app/core/models/sheet-model.ts
```diff
@@ -565,15 +570,34 @@ export class SheetModel {
...
     /**
-     * Set whether the label of the sheet should be highlighted or not
+     * Set whether the label of the sheet should be labelHighlighted or not
      * @param val enable disable highlight
      */
-    setHighlight(val: boolean) {
-        if (val !== this.highlighted) {
-            this.label.material.map =
-                this.generateLabelTexture(this.id.toString(), val ? this.labelHighlightColor : this.color);
+    setLabelHighlight(val: boolean) {
```
Chessnl: This function can and should be covered with test-cases. This (in my opinion) does not necessarily need to be done in this pull request. But if it is not, do make an issue to implement it.

src/app/core/models/sheet-model.ts
```diff
@@ -565,15 +570,34 @@ export class SheetModel {
...
-            this.highlighted = val;
+            this.labelHighlighted = val;
+        }
+    }
+
+    /**
+     * Set whether the sheet is sheetHighlighted or not
+     * @param val enable disable highlight
+     */
+    setSheetHighlight(val: boolean) {
```
Chessnl: This function can and should be covered with test-cases. This (in my opinion) does not necessarily need to be done in this pull request. But if it is not, do make an issue to implement it.

src/app/workspace/viewport/viewport.component.ts
```diff
@@ -25,27 +27,9 @@ import { TimingFileType } from '../../import/timing-file-type.enum';
 export class ViewportComponent implements AfterViewInit, OnDestroy {
 
     /**
-     * Color sets for multiple timings.
+     * Maximum difference allowed between 2 timing states positions before being considered an error
```
Chessnl:
```suggestion
     * Maximum difference allowed between 2 timing states positions before being considered a significant deviation
```

I prefer this more because an Error might make it seem that it is an error within our program. However, we wish to indicate that there is a difference between expected and real (which might potentially be caused by an error in the measured system.)

src/app/workspace/viewport/viewport.component.ts
```diff
@@ -127,9 +122,19 @@ export class ViewportComponent implements AfterViewInit, OnDestroy {
     private stats: STATS;
 
     /**
-     * Map of sheet models. First index is sheet id, second index is timing id.
+     * Raycaster to detect which object the user is looking at
+     */
+    private rayCast: RayCast;
+
+    /**
+     * The hit of the RayCast
```
Chessnl:
```suggestion
     * A sheet hit by a RayCast
```

src/app/workspace/viewport/viewport.component.ts
```diff
@@ -274,17 +215,57 @@ export class ViewportComponent implements AfterViewInit, OnDestroy {
...
+        this.rayCast = new RayCast();
         // Create a Perspective Camera
         this.camera.initializeCanvas(this.canvasRef.nativeElement);
     }
 
+    /**
+     * Mouse Event in canvas
+     * @param event has contents of the event
+     */
+    mouseMove(event: MouseEvent): void {
```
Chessnl: The doc of this comment should be more extensive.
E.g. A Mouse Event is made. Update the mouse coordinates to its coordinates in the canvas local coordinate system.
@param event contains the new coordinates of the mouse

Chessnl: This function can and should be covered with test-cases. This (in my opinion) does not necessarily need to be done in this pull request. But if it is not, do make an issue to implement it.

Chessnl: If this one is however not easy to cover, ignore this comment.

src/app/workspace/viewport/viewport.component.ts
```diff
@@ -375,16 +357,34 @@ export class ViewportComponent implements AfterViewInit, OnDestroy {
                                 0, // default value
                                 this.route.getSegmentPosition(sheetTimingState.getPosition()).areSheetsUpsideDown()
                             );
+
+                            // Check if sheet should be highlighted
```
Chessnl: Check whether it is possible to have a test covering these lines of code. If it is not, ignore this comment.

src/app/workspace/viewport/viewport.component.ts
Erik9519: Addressed

src/app/workspace/viewport/viewport.component.ts
Erik9519: Addressed

src/app/workspace/viewport/viewport.component.ts
Erik9519: Could be tested but would require a fairly extensive test case as the sheet state engine would also have to be in working order to verify it.

src/app/workspace/viewport/viewport.component.ts
Erik9519: Might be difficult to test as the size of the ui canvas would have to be known in order to verify the result and the size of it is not defined but rather is manually set through resize events.

Erik9519: Addressed

src/app/core/models/sheet-model.ts
Erik9519: Could be done bu manually checking the material colors to verify that it works however I'd address in a separate issue

src/app/core/models/sheet-model.ts
Erik9519: Difficult to verify the result as the only thing that's publicly available to verify that it works is the texture of the label itself but not the actual color of it.

src/app/core/models/sheet-model.ts
Chessnl: Tag #453

PR #465, opened 2019-06-17
#434 Sheet names are now generated using the type of the timing file.

TijnBertens:
#### What have I done
Timing files are now associated with a type (REAL or EXPECTED). The names of sheets are now generated using these file types. In the process, the UI for the import screen was adapted as well.

#### Testing
Unit testing.
Run the application and inspect the input screen, table, and graph. 

#### Addresses SRD section(s)
Import
Parser Service
Timing Parser
Sheet Service
Sheet Container


src/app/core/models/sheet/sheet-container.ts
```diff
@@ -43,19 +44,33 @@ export class SheetContainer {
...
+     * @param timingFileType The type to add
+     */
+    addTimingFileType(timingFileType: TimingFileType): void {
+        this.timingFileTypes.push(timingFileType);
     }
 
     /**
      * Clears the data in the container
      */
     clear() {
```
Erik9519: Missing return type

TijnBertens: Fixed

PR #472, opened 2019-06-18
Removed distance to previous poi column

nolmoonen:
#### What have I done
Removed distance to previous poi column. Code currently in place was work-around and relied on error catching. Feature will not be implemented due to time constraints.

#### Testing
None.

#### Addresses SRD section(s)
None.


Chessnl:
Hihi nee. Zit in de URD schat :-)

Issue #470, opened 2019-06-18
Code duplication resolve


Chessnl:
Automatic testing gives code duplication. Within this issue, the amount of code duplication on development should be reduced to a minimum.

PR #473, opened 2019-06-18
#470 Code duplication resolve

Chessnl:
#### What have I done
Addressed issue #470 and immediately added additional test cases for the UTP for the column and row

#### Testing
Check whether everything still works as before

#### Addresses SRD section(s)
None


Dacaspex:
Conflicts

PR #476, opened 2019-06-18
Bugfix/stop dispatcher during parsing

### LINKED ISSUE #193: Refactor import.component.ts is*() functions to reduce repeated code
The code below can be simplified into one function, as well as probably a couple of other methods
```
    /**
     * Check if the ui state is import manual
     */
    isWelcome() {
        return this.startState === StartState.Welcome;
    }

    /**
     * Check if the ui state is import manual
     */
    isImport() {
        return this.startState === StartState.Import;
    }

    /**
     * Check if the ui state is import manual
     */
    isImportManual() {
        return this.startState === StartState.ImportManual;
    }
```

TijnBertens:
#### What have I done
Fixes #193.

Fixed the bug pointed out by Cas. When first importing a single timing file and then importing two timing files, an error would occur where the sheet state engine tries to update sheet timing states that do not exist. This was due to the SheetContainer being updated during parsing before the SheetService was updated.

To address this, the dispatcher is now disabled during parsing. Hence, the sheet state engine will not update.

This pull request also includes some clean-up in the import component. The number of function is now reduced.

#### Testing
Unit tests.
Try reproducing the bug

#### Addresses SRD section(s)
Import Component


Dacaspex:
Conflicts

Issue #433, opened 2019-06-13
Two circles at start import page, numbers are off-center.


Astrejoe:
![off_center](https://user-images.githubusercontent.com/10343246/59447025-e5eb6980-8e02-11e9-9010-e4bccc69a540.png)

PR #487, opened 2019-06-19
Centers the numbers in the two circle icons on the start of the impor…

Astrejoe:
#### What have I done
Fixes issue #433. I centered the two numbers in the circles on the start import screen. This was mainly an issue on firefox. 

#### Testing
Open the application on both firefox and chrome and verify that the two circles on the start import screen look good.

#### Addresses SRD section(s)
/


ArtixOfVenion:
The code looks good and the approach seems correct however when I looked at it on my pc I still felt that it is somewhat off-center. I'm not sure if its meant be grounded at the bottom of the textbox but right now the number is grounded at the bottom making it such that the number is a bit below the center. See the image below:
image

Astrejoe:
> The code looks good and the approach seems correct however when I looked at it on my pc I still felt that it is somewhat off-center. I'm not sure if its meant be grounded at the bottom of the textbox but right now the number is grounded at the bottom making it such that the number is a bit below the center. See the image below:
> image

This is now fixed.

